PR #93, opened 2013-04-29
Update daySelectionMousedown to interpret OSX CTRL + Click as a right mouse click

spinegar:
When using CTRL + Click in OSX (right click) the selectable feature clears out the currently selected days. Adding !ev.ctrlKey prevents FullCalendar from clearing out the selected days when using CTRL + Click in OSX.

### DIFF
--- a/src/common/SelectionManager.js
+++ b/src/common/SelectionManager.js
@@ -1,4 +1,3 @@
-
 //BUG: unselect needs to be triggered when events are dragged+dropped
 
 function SelectionManager() {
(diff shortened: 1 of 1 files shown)


src/common/SelectionManager.js
```diff
@@ -69,7 +68,7 @@ function SelectionManager() {
 		var cellIsAllDay = t.cellIsAllDay;
 		var hoverListener = t.getHoverListener();
 		var reportDayClick = t.reportDayClick; // this is hacky and sort of weird
-		if (ev.which == 1 && opt('selectable')) { // which==1 means left mouse button
+		if (ev.which == 1 && opt('selectable') && !ev.ctrlKey) { // which==1 means left mouse button, ctrlKey + left mouse button = osx right click
```
SudharakaP: I think it might be better to modify this if statement to include a check for the platform since this change will disable ctrl click on windows. How about checking for the windows platform as follows? 

if (ev.which == 1 && opt('selectable') && (!ev.ctrlKey || window.navigator.userAgent.includes("Windows"))){...}

SudharakaP:
@arshaw , @spinegar : I was wondering that it would be better to include a check for the platform since this check would disable ctrl select on windows. How about changing the if statement as follows? 

`if (ev.which == 1 && opt('selectable') && (!ev.ctrlKey || window.navigator.userAgent.includes("Windows")))`

PR #99, opened 2013-05-21
Fully working horizontal resource views for Fullcalendar

jarnokurlin:
Fully working horizontal resource views for Fullcalendar. You can test it in http://tux.fi/~jarnok/fullcalendar-resourceviews

Hope you like it.


csellis:
Is there a reason why my AVG is detecting a virus with this?

rajgoraya:
Awesome work Jarno! I am using your branch and it has worked out great for us. 
Is there some way to control the top down order within the same slot or within overlapping slot range. In the image below I was able to change the order of first by reordering the event list. I am trying to figure out how to move the 3rd one on top since it starts early then others. 
In short I am trying to control the top o down order so that the slot starting early may always be top and some way to reorder the slots ones starting at same time. I will appreciate if some can point me in the right direction. 
capture

arshaw:
If you are using @jarnokurlin's implementation, I encourage you to keep using it, and I encourage him to continue hosting that fork, but from the fullcalendar project's perspective, for all intents and purposes, the official implementation of this feature has been released here:
http://fullcalendar.io/scheduler/

(NOTE: it is released under different terms than FullCalendar's core)
info about the release:
http://blog.arshaw.com/1/post/2015/08/scheduler-released.html

PR #179, opened 2014-07-11
Remove bower install from npm install script

timmywil:
For those of us using npm as the front-end package manager, forcing a bower install throws a wrench in the gears. I'd much rather be responsible for installing these dependencies myself.


arshaw:
released.
https://github.com/arshaw/fullcalendar/releases/tag/v2.3.1

PR #180, opened 2014-08-06
Added Uyghur language

boyaq:
Hi,

Thanks for the great library. Here is Uyghur translation of select2, hope you can merge it. Thanks a lot.

Best,

Alim

### DIFF
--- a/lang/ug-cn.js
+++ b/lang/ug-cn.js
@@ -1,0 +1,10 @@
+
+$.fullCalendar.lang("ar", {
+	defaultButtonText: {
+		month: "ئاي",
+		week: "ھەپتە",
+		day: "كۈن",
+		list: "كۈنتەرتىپ"
+	},
+	allDayText: "پۈتۈن كۈن"
+});


arshaw:
note to self: figure out "eventLimitText" text

lang/ug-cn.js
```diff
@@ -0,0 +1,10 @@
+
+$.fullCalendar.lang("ar", {
```
arshaw: should the language code be `"ug-cn"` here?

arshaw:
I would love to include this translation, but a prerequisite is that MomentJS must also support it. [Look a MomentJS's locale list](https://github.com/moment/moment/tree/develop/locale), it seems they do not support Uyghur yet. Am I looking for the wrong language code (is it `ug-cn`)? If MomentJS simply doesn't have it translated yet, could you contribute to their project?

caseyjhol:
See https://github.com/moment/moment/pull/1834

arshaw:
this can now be ported over to v4. i'll do that for the next release.

arshaw:
this is released in v4.4.0:

I renamed this to `ug`

HOWEVER, the browser must support the `ug` locale in order for this to work. On my computer Firefox does, but Chrome doesn't for example. If you want more consistent support, it's advised to use the moment plugin.

PR #195, opened 2014-09-30
Vertical resource view (churchdesk)

zvictor:
Long waited resource view implementation, now for fullcalendar 2.2 and based on @seankenny work.

Merge it, compile it with grunt, and open `demos/resource-views.html` to see it working.
There is also a fixed columns width version at `demos/resource-fixed-views.html` (cannot be run locally), but this one is just a guide of how you can implement it by yourself on your installation.

-------
screenshot

--------
resourceview2


atothewest:
This is fantastic work but one critical thing is stopping me switching from @seankenny's version: 

When dropping events, calling 'revertFunc' returns an event to the original time, but not to the original resource.

davidwoods:
Can we expect this to be merged soon, its a very popular request?

riula:
@zvictor great work!!!

davidwoods:
@zvictor is their anyway to get the resourceid when using the dayClick event? BTW big thank you for your work on this!

zvictor:
@arshaw any comments on this?

arshaw:
Hi @zvictor, looks like some great work. I'm going to download this and play with it. I'm very busy working on [Timeline View](http://blog.arshaw.com/1/post/2015/02/timeline-and-other-news.html) and I'm not completely sure how this vertical resource view will fit in with it, since they are both resource-related features. I'm a bit hesitant to merge it in right now, because even though I'm sure you would be very helpful in maintaining it post-merge, the default expectation becomes that the core project (FullCalendar) take responsibility over the feature and the resulting related feature requests, which I'm sure there will be a lot of since people's usecases for "resources" are extremely varied. I just remember when I released the initial "agenda" view, how many new feature requests were created and ultimately how much more work the project became. So, as of right now, I'm more apt to work with authors of bulky new features like this in the context of external plugins rather than as part of the core. Please feel free to PM me.

rasmuschristensen:
@zvictor  This works really great. What is the best approach to reload events and replace already loaded events. I've tried using removeEventSource, and also removeEvents, where I used the resourceId as a filter. Both approaches makes the calendar total blank. The show/hide of a resource works great, but I need to replace the events when a resource becomes visible.

Keep up the great work!

marronsuisse:
@zvictor and for anyone considering using the ChurchDesk fullcalendar fork, it seems dropping external events is not working. Let me know if I'm wrong.

T2theC:
AWESOME. Thanks for this. I have been using Sean's version for a while and it is a little outdated now.

I've been going over the code and can't seem to figure this out:

Is there a way to refetch resources? Similar to `$('#calendar').fullCalendar( 'refetchEvents');` ?

My resources are people with tasks as events. One week there maybe a `person` that has `10` tasks. I can filter the resource to match this. However, when I move to the next week, the `person` in question has no tasks and a different `person` has `5` tasks. I'd like to refetch the resource when the date changes so I can reorder the resources.

Any ideas how to do this? I am hacking away and should be able to pass in the current date and append it to the resource as a query string. So I'd like then to refetch the source.

Many thanks.

goodkiller:
Please merge this pull!

lasterra:
@zvictor It looks that Resource view did not render background events

lasterra:
@zvictor Sorry, resource view render background event when has resources asigned to it. It works perfectly for me.

espen:
I think this view is a great idea and looking forward to getting to use it. But I really think this project should have a discussion about plugins. I see now the latest commits introduces plugin support. I think this resource view should be done as a plugin instead of adding more code to the project. While some of us are using Fullcalendar within a fully featured app others are displaying a simple calendar on their public website and thus code size is more important and extra features such as this is not so important.

rzane:
I agree with @espen. Because @arshaw said he doesn't plan to merge this, having it as a plugin would make it easier to maintain. I was using @jarnokurlin's fork from #99, but because it was never merged, it completely died off and stopped being updated. I'd hate to see the same happen here.

zvictor:
Thank you all for your interest on this project. I see a good discussion going on here and many important issues have been raised, but I think it is the time to consider closing this PR. It's creation date was almost one year ago, and since then few changes have been added or updated.

I like to see that things are heading more to the 'plugin' way nowadays than it was before, and I hope in the future more features like the proposed on this PR are gonna be straightforward to implement.

The resource view still works well for most of the cases and can be checked out at [ChurchDesk/fullcalendar](http://github.com/ChurchDesk/fullcalendar), but it is not updated with the main fullcalendar fork anymore.

arshaw:
@zvictor I'm going to close this out. Thanks for helping everyone out with all your great work.

The FullCalendar project has created a separate implementation of this:
http://fullcalendar.io/scheduler/
http://fullcalendar.io/docs/vertical_resource_view/getting_started/
http://fullcalendar.io/blog/2015/02/timeline-and-other-news/ (old blog post)

zvictor:
Thank you @arshaw.

@andreiho be aware =)

PR #204, opened 2014-12-05
Sorted events for renderMoreLink

bio:


arshaw:
I made an alternate fix in 7772869c8bdde1b0780442468286e864a61d55d1, it covers both `reslicedAllSegs` and `reslicedHiddenSegs`. Thanks!

PR #205, opened 2014-12-12
Support installation via npm

dansomething:
This follows PR #179 to remove `bower install` and specifies the files that need to be made available in the published artifact so this project can be properly installed via npm.


arshaw:
released
https://github.com/arshaw/fullcalendar/releases/tag/v2.3.1

PR #206, opened 2014-12-19
Added option to set the calendar height to 'parent'

marcucio:
Added option to set the calendar height to 'parent' where the calendar will take the height of the parent container. This is useful if you would like the calendar to just fill a div in the window.


arshaw:
Would you mind creating a new issue tracker ticket as described in this article?

https://github.com/arshaw/fullcalendar/wiki/Contributing-Features

Thanks so much!

marcucio:
no problem, added here:

https://code.google.com/p/fullcalendar/issues/detail?id=2439

arshaw:
Thank you for creating that and linking to the other PR as well. I will be closing this PR in order to keep the PR list groomed, but will come back and reference this when the times comes to integrate your code and fix the bug. Thanks.

PR #209, opened 2014-12-30
fix bug when trying to close popover in month view

marcucio:
When the more events popover is open and you click on a different day in month view the day click event was being fired, we should just close the popover and not fire additional events.


arshaw:
Would you mind creating an issue tracker issue for this [as described here](https://github.com/arshaw/fullcalendar/wiki/Contributing-Features)? Thanks so much!

marcucio:
here you go:

https://code.google.com/p/fullcalendar/issues/detail?id=2440

arshaw:
Thank you! I'll be closing this PR in order to keep the PR list cleaner, but will come back and reference this PR when it is time integrate your code and fix the bug. Thanks!

PR #211, opened 2015-01-05
Add event sources getter

antonioribeiro:
This was just to add

	t.getEventSources = getEventSources;

and

	function getEventSources() {
		return sources;
	}

But there were a bunch of lost tabs in the source.


arshaw:
Would you mind creating an issue tracker ticket [as described in this article](https://github.com/arshaw/fullcalendar/wiki/Contributing-Features). Thanks so much!

PR #214, opened 2015-01-08
change timeMin and timeMax params to use ISO date

mjatharvest:
Google Calendar API v3 requires timeMin and timeMax to be formatted as yyyy-mm-ddT00:00:00+00:00 or sometime you will get a 400 bad request error

### DIFF
--- a/src/gcal/gcal.js
+++ b/src/gcal/gcal.js
@@ -113,8 +113,8 @@ function transformOptions(sourceOptions, start, end, timezone, calendar) {
 
 	data = $.extend({}, sourceOptions.data || {}, {
 		key: apiKey,
-		timeMin: start.format(),
-		timeMax: end.format(),
+		timeMin: start.format("YYYY-MM-DD") + 'T' + '00:00:00+00:00',
+		timeMax: end.format("YYYY-MM-DD") + 'T' + '00:00:00+00:00',
 		timeZone: timezoneArg,
 		singleEvents: true,
 		maxResults: 9999


espen:
Could put everything in the moment format: format("YYYY-MM-DDT00:00:00+00:00')

mjatharvest:
Thanks for the feedback. I tried using format() only, but found that the query params didn't work quite right. Notice how there is no "T" in the urls below.

Using Moment format("YYYY-MM-DDT00:00:00+00:00') I get
&timeMin=2014-12-27A00%3A00%3A00%2B00%3A00&timeMax=2015-02-09A00%3A00%3A00%2B00%3A00

Concatenating strings format("YYYY-MM-DD") + 'T' + '00:00:00+00:00' I get
&timeMin=2014-12-27T00%3A00%3A00%2B00%3A00&timeMax=2015-02-09T00%3A00%3A00%2B00%3A00

I'm not sure if that is an issue with moment or with jquery's ajax data. But the request to Google Calendar  API fails without the "T"

espen:
Sounds strange.

  end.format("YYYY-MM-DDT00:00:00+00:00") == end.format("YYYY-MM-DD") + 'T' + '00:00:00+00:00'
  true

arshaw:
@mjatharvest this scenario should already be covered. [See my comments in the code here](https://github.com/arshaw/fullcalendar/blob/master/src/gcal/gcal.js#L101).

(NOTE: if the moment doesn't have a zone it also doesn't have a time)

If you can prove there is somehow a bug, I will look at this more closely.

PR #215, opened 2015-01-10
introduce auto-height to calculate the height based on parent element

chris-rock:
This enables fullcalender to derive the height of the grid from the parent height.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -505,7 +505,7 @@ function Calendar(element, instanceOptions) {
 
 
 	t.isHeightAuto = function() {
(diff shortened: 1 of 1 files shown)


samedii:
Same as https://github.com/arshaw/fullcalendar/pull/206 ?

chris-rock:
I think #206 would work as well.

arshaw:
See [my comment here](https://github.com/arshaw/fullcalendar/pull/206#issuecomment-73764787).. closing this issue, but will reference this ticket when the time comes. [Associated issue tracker ticket](https://code.google.com/p/fullcalendar/issues/detail?id=2439). Thanks for your work!

PR #216, opened 2015-01-12
Latin calendar fix

megajpc:
In Portugal (and latin derived speaking countries) the month names are capitalized


espen:
You are editing the dist files. Not the source itself. If you want to change the locale settings for Fullcalendar check https://github.com/arshaw/fullcalendar/blob/master/lang/es.js

Also Fullcalendar uses Moment for displaying dates. So to specify how Moment should display month name (or make a pull request to Moment) please see http://momentjs.com

arshaw:
Thanks! But could you please post another commit and PR with a non-minified version of the translation file? Something like one of these files:

https://github.com/arshaw/fullcalendar/tree/master/lang

PS- you won't need to include the moment or datepicker translations.

PR #217, opened 2015-01-12
added more translations to Norwegian locale

espen:


arshaw:
im just now getting around to merging all these i18n PR's. really sorry for taking forever.

seems that all the latest date libs break up into norwegian nynorsk (nn) and norwegian bokmål (nb).

could you incorporate these changes into one or the other (or both?). i'll close this PR and await another.

also, i'm hoping we can maintain consistency w/ start each word with capital letters.

the generated fc translations are an amalgamation of [jquery-ui's](https://github.com/jquery/jquery-ui/tree/master/ui/i18n) and [moment's](https://github.com/moment/moment/tree/develop/locale). Could you please look at those libs to verify that your changes are not now redundant?

PR #218, opened 2015-01-22
merge base in

theshaun:


theshaun:
my bad, was pulling the wrong way

PR #219, opened 2015-02-06
Update fr.js

Reeslo:
add traduction for today object.


arshaw:
Hmm, fullcalendar already has this translation. It actually comes from the jQuery datepicker translations it leverages behind the scenes. See [this demo](http://fullcalendar.io/js/fullcalendar-2.2.6/demos/languages.html) (choose fr at the top) and you will see it working.

PR #220, opened 2015-02-24
Fix week view column title for en-gb

gidoca:
Since Fullcalendar 2.3, the display of the column header in the week views is broken when the language is set to en-gb. Instead of the weekday and date, it just displays "object Object", see [this js-fiddle](http://jsfiddle.net/q1a40bno/2/).

### DIFF
--- a/lang/en-gb.js
+++ b/lang/en-gb.js
@@ -1,6 +1,8 @@
 
 $.fullCalendar.lang("en-gb", {
-	columnFormat: {
-		week: 'ddd D/M'
+	views: {
+		week: {
+			columnFormat: 'ddd D/M'
+		}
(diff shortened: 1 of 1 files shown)


arshaw:
Gaa, this is embarrassing. Thanks for the fix.

PS- I'm not actually sure if lang files can specify a `views` option. What you are seeing might be the date format falling back to the one being computed off of moment's settings (which might be good enough actually). I forget why I overrode en-bg specifically.

gidoca:
You are right, it works perfectly fine without the option.

arshaw:
fixed https://github.com/arshaw/fullcalendar/releases/tag/v2.3.1

PR #222, opened 2015-02-27
CommonJS support added

jesusprubio:
[UMD](https://github.com/umdjs/umd) support added to be compatible with Browserify (and others), like they recommend [here](https://github.com/umdjs/umd/blob/master/jqueryPluginCommonjs.js) for JQuery plugins. Maybe it would be good to add something like this to the documentation.

This library is [CommonJS](http://www.commonjs.org/) compatible, so you can use it in this way:

```javascript
var jquery = require('jquery'),
    moment = require('moment'),
    fullcalendar = require('fullcalendar')(jquery, moment);

    jquery('#calendar').fullCalendar({
          weekends: false // will hide Saturdays and Sundays
    });
    ...
```

### DIFF
--- a/dist/fullcalendar.js
+++ b/dist/fullcalendar.js
@@ -6,9 +6,13 @@
 
 (function(factory) {
 	if (typeof define === 'function' && define.amd) {
+        // AMD, register as an anonymous module
 		define([ 'jquery', 'moment' ], factory);
-	}
-	else {
+	} else if (typeof exports === 'object') {
+        // Node/CommonJS
+        module.exports = factory(require('jquery'), require('moment'));
+    } else {
+        // Browser globals
 		factory(jQuery, moment);
 	}
 })(function($, moment) {


jesusprubio:
Really this PR is not needed, but I think you need to generate the distribution file again. Differences:
intro.js
```
(function(factory) {
	if (typeof define === 'function' && define.amd) {
		define([ 'jquery', 'moment' ], factory);
	}
	else if (typeof exports === 'object') { // Node/CommonJS
		module.exports = factory(require('jquery'), require('moment'));
	}
	else {
		factory(jQuery, moment);
	}
})(function($, moment) {
```
dist/fullcalendar.js
```
(function(factory) {
	if (typeof define === 'function' && define.amd) {
		define([ 'jquery', 'moment' ], factory);
	}
	else {
		factory(jQuery, moment);
	}
})(function($, moment) {
```

arshaw:
Thanks for the PR. The dist files will be re-generated when the next version is released. I will also submit this to the NPM registry after that happens.

jesusprubio:
Nice :), thank you

PR #223, opened 2015-03-05
	fix view show bug that in Zh-cn view

struCoder:
one problem that in Zh-cn model
when i click day|month|week button the prev-button and next-button does not change
and the correct is  
qq20150306-2
qq20150306-3
qq20150306-4

maybe you have good codes : )


struCoder:
sorry, just now i found that is a common problem i will try to fix

arshaw:
sounds good

PR #224, opened 2015-03-13
Add option autoUpdate fixing issue 2293

diogoalexandrefranco:
Add option that forces calendar to fetch and render events every x
seconds.

I don't know if this is mergeable but since i, like others, were doing refetchEvents() on a setInterval, i thought to do this so that we can simply use a calendar option and so that event's don't flicker on update.

Thanks for this awesome plugin.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -415,6 +415,7 @@ function Calendar_constructor(element, overrides) {
 	var ignoreWindowResize = 0;
 	var date;
 	var events = [];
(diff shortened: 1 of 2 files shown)


arshaw:
Hmm, I'm gonna close this PR per the [feature request guidelines](https://github.com/arshaw/fullcalendar/wiki/Contributing-Features), but can you please make a ticket on the issue tracker in order to gauge interest in this feature?

PR #225, opened 2015-03-17
Update hr.js

hubgit3:
Fixed several typos


arshaw:
Thanks! But could you please post another commit and PR with a non-minified version of the translation file? Something like one of these files:

https://github.com/arshaw/fullcalendar/tree/master/lang

PS- you won't need to include the moment or datepicker translations.

PR #226, opened 2015-03-20
Regular slot duration is a multiple of 5

thiagotalma:
I understand that a normal slot should be a multiple of 5.
It makes no sense to use "slotDuration = 20" and display all times.

If you disagree with the PR, please make this rule a configuration parameter.

Thanks and hugs!

### DIFF
--- a/src/common/TimeGrid.js
+++ b/src/common/TimeGrid.js
@@ -71,7 +71,7 @@ var TimeGrid = Grid.extend({
 		var view = this.view;
 		var isRTL = this.isRTL;
 		var html = '';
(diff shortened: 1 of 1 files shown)


arshaw:
Closing this PR per the [feature contribution guidelines](https://github.com/arshaw/fullcalendar/wiki/Contributing-Features) but will link to this PR from the original [issue tracker ticket](https://code.google.com/p/fullcalendar/issues/detail?id=946).

arshaw:
Actually, will not link to it, as it contains a new hardcoded value that some ppl might not like (as you know), but hope you are glad to hear that myself and others are aware of the shortcoming and keeping track of it on that ticket.

PR #227, opened 2015-03-28
Update

jdwisse:
In version 2.1.1 there is a bug when you update an allDay event. The end is always lost. In the new version 2.3.1 this is solved. I try to merge your code with the new 2.3.1 version but it is not compatible due some changes. Do you have plans to update your code in the near future or shall i give it a try? Thanks in advanced...


espen:
If the bug is fixed in version 2.3.1 then why not just use that version? What exactly have you included in this mega pull request? Are these new bugfixes or why all these changes?

PR #228, opened 2015-03-31
Include /dist/lang when installing via bower

joernv:


arshaw:
hmm, the langs *should* be included in the lang folder when you do a bower install. That change you made actually prevents them from being included. (`*` means don't traverse into subdirectories)

PR #229, opened 2015-04-29
Added eventDropOutside functionality

JamShady:
Allows events to be captured when an event is dragged outside the calendar onto an empty area


JamShady:
Closed - repo got nuked because it was linked to something else

PR #231, opened 2015-05-12
add week to similarUnitMap used for formatRange (also fixes issue 2371)

r-springer:
fixes the problem stated here: https://code.google.com/p/fullcalendar/issues/detail?id=2371&q=duplicate%20title&colspec=ID%20Type%20Status%20Milestone%20Summary%20Stars


arshaw:
I will close this PR for now but will leverage it when #2636 is addressed

PR #235, opened 2015-06-25
Fix MouseFollower stop animation compete callback

redexp:
Issue https://code.google.com/p/fullcalendar/issues/detail?id=2529


arshaw:
This is a fix for #2794. I will close this PR for now but will refer back to it and leverage it when it is time for that issue to be addressed.

PR #238, opened 2015-06-30
Arbitrary differing slotDuration increments (fixes issue 201)

kleisauke:
Implementation of issue [201](https://code.google.com/p/fullcalendar/issues/detail?id=201). You can test it here:
http://kleisauke.nl/fullcalendar/demos/timeslots-views.html

If you not want to snap on the timeslot while dragging/resizing (`snapDuration` is set to 5 minutes):
http://kleisauke.nl/fullcalendar/demos/timeslots-views-without-snap.html

I've build this because I'm writing a scheduler for a school system that has varied timeslots. It's tested it in the week and day views and it works there.

Timeslots can be set using: `slots: [{start: '07:35', end: '08:20'},{start: '08:25', end: '09:10'}]`. It automatically support breaks.
There's one new `snapOnSlots` option added, if this option is set to true then it snaps on the timeslot while dragging/resizing. If it's set to false then it snaps not on the timeslot but it snaps to the duration what is set in `snapDuration` (default half an hour).

I hope you'll like it.


danielbsig:
+1 for incorporating this into the library.

RicardoEmpre:
This is great! I hope is pulled into the calendar, I have to use something like it in the next weeks!

benoitmasson:
Thanks for the work, I really needed that!

I've made a pull request (https://github.com/kleisauke/fullcalendar/pull/1) from your PR to extend it with some more functionalities that were necessary for me too. I'm detailing them here in case anybody else needs that:
* **Make code more robust**: added parsing of slots in `processOptions()`, to check and filter valid slots, sort them and store them in the main object
* **Add a `showSlotEndTime` option**: when set to `false`, the end time of the slots is not displayed
* **Add `minor` slots**: slots can have a `minor` attribute, which, when set to true, renders them as minor slots (dotted horizontal border). Option `showMinorSlotTime` set to `false` prevents time display on minor slots.
* **Add a `className` attribute to slots**:  custom slots can have a `className` attribute, containing class names that will be added to the corresponding `<tr>` element.
* **Simplified and split code for `computeTimeTop()`**: split code into sub-functions; factorized and commented some `if-then-else` cases to make them more intuitive; replaced `this.breakHeights` by `this.slatBottoms`
* **Extend `snapOnSlots` with a few options**: `snapOnSlots` can be `true` or an object:
```javascript
    {
      snapEffectiveDuration: false, // default: false
      snapPolicy: 'enlarge'         // default: 'enlarge', could also be 'closest'
    }
```
* 
  * If `snapOnSlots.snapEffectiveDuration === true`, then the new duration after dragging an event will try to preserve the initial effective duration (duration without breaks).
  * If `snapOnSlots.snapPolicy === 'enlarge'`, if the border of a dragged or resized event falls in the middle of a slot, its duration is extended to fill the entire slot.
        If `snapOnSlots.snapPolicy === 'closest'`, the border is moved to the closest timestamp marking the border of a slot.

arshaw:
I am closing this issue for organizational reasons, but will leverage it when tackling #472 (the github issue version of the original)

PR #256, opened 2015-08-04
Added a 'refetchBgEvents' method to refetch only background events

### LINKED ISSUE #254: refetchEvents for multiple specific sources
Hi,

I'm using both regular and background events on my calendar. Background events are used to highlight forbidden places for regular events, and are updated dynamically whenever such an event is selected (`eventClick()` or `eventDragStart()` callbacks) via the method '`refetchEvents`'.

In the case of a click, everything works well. In the case of a drag however, the '`refetchEvents`' causes the selected event to be refetched as well, thus leading to graphical glitches (event displayed twice, initial event not moving, ...).

The best workaround I found for this issue is to set a global property to prevent that the event is displayed (in the `eventRender()` callback) while it is dragged, and then in the `eventDrop()` callback I can manually update its start and end times according to delta. This is a ugly hack and it doesn't solve all bugs in my application.

A much simpler solution to that would be to allow to refetch events from a single source, leaving other events unchanged, but I don't think this is possible with the current API.

Is there something I am missing?

Thanks in advance

benoitmasson:
When called, the method 'refetchBgEvents' clears, refetches and renders only background
events (rendering = 'background' or 'inverse-background'), leaving "regular"
events unchanged.

Useful when dragging events changes the background: if 'refetchEvents' is
called, then the currently dragged event is incorrectly rendered since its
source 'event' object has been invalidated by the refetch.

That should resolve #254


arshaw:
this solution might be a little too narrow (should be robust enough to handle any criteria, not just bg events) but i see the need for it

arshaw:
closing this, but i will reference it when tackling #254

PR #257, opened 2015-08-04
Option to set a day break time for TimeGrid

matiaslarsson:
Accepted in: https://code.google.com/p/fullcalendar/issues/detail?id=2541

Setting the new option dayBreakTime to for example "07:00" or "7:00 AM" will make the day columns in TimeGrid views start at 7 am in the morning and extend past midnight, ending 7 am the following day. This can be useful for TV-listings or concert/club event scheduling were it's common to have "days" ending past midnight.

This is accomplished with two small code changes.

1) Make sure we actually use the seg column date as the startOfDayDate property in computeSegVerticals() instead of the seg event start date.

2) Add a dayBreakTime option which overrides the minTime and maxTime values to make the columns render past midnight.

Example;

$('#calendar').fullCalendar({
   header: {
      left: 'prev,next today',
      center: 'title',
      right: 'month,agendaWeek,agendaDay'
   },
   defaultView: 'agendaWeek',
   dayBreakTime: "07:00",
   ...
});

### DIFF
--- a/src/common/TimeGrid.events.js
+++ b/src/common/TimeGrid.events.js
@@ -93,8 +93,8 @@ TimeGrid.mixin({
 
 		for (i = 0; i < segs.length; i++) {
 			seg = segs[i];
-			seg.top = this.computeDateTop(seg.start, seg.start);
-			seg.bottom = this.computeDateTop(seg.end, seg.start);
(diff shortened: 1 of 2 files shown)


arshaw:
closing this, but will use this PR when #2806 is addressed

PR #258, opened 2015-08-04
Firing event-click for bg events

matiaslarsson:
Accepted in: https://code.google.com/p/fullcalendar/issues/detail?id=2543

In some cases it's useful to detect clicks on background events (not currently supported).

The small fix is divided into two parts;

1) Make sure we connect the seg to the fillSeg by fc-seg data in renderFillSegEls() in Grid.js, just like we do for ordinary event segs.

2) Detect click events on .fc-bgevent-container elements and trigger bgEventClick in a new bindBgSegHandlers method.

We could extend the bindSegHandlers method to support bg events too, but the new method felt cleaner since we don't want to resize the background events.

Example;

$('#calendar').fullCalendar({
   header: {
      left: 'prev,next today',
      center: 'title',
      right: 'month,agendaWeek,agendaDay'
   },
   defaultView: 'agendaWeek',
   bgEventClick: function(calEvent, jsEvent, view) {
      alert(calEvent.title);	
   },
   ...
});

### DIFF
--- a/src/common/Grid.events.js
+++ b/src/common/Grid.events.js
@@ -215,6 +215,30 @@ Grid.mixin({
 			}
 		);
 	},
+	
+	bindBgSegHandlers: function() {
+		var _this = this;
+		var view = this.view;
+
+		$.each(
+			{
+				click: function(seg, ev) {
+					return view.trigger('bgEventClick', this, seg.event, ev); // can return `false` to cancel
+				}
+			},
+			function(name, func) {
+				// attach the handler to the container element and only listen for real event elements via bubbling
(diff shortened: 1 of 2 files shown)


arshaw:
will potentially leverage this when issue #2808 is addressed

nesbittg:
I implemented this but the defaultView: 'agendaDay',
            bgEventClick: function (calEvent, jsEvent, view) {
                alert(calEvent.title);
            },       
was not being triggered.  I looked at the code and handleSegClick was being triggered but it didn't bubble back up the the actually client event.  Is there something I need to do to get this to work?  Thx!

PR #2833, opened 2015-08-20
eventLimitRender option

rumbu:
Added an option `eventLimitRender` to have an ability to manage the 
rendering of "more" links like it is for events using `eventRender` 
option. Docs will be in the branch docs-2.0, separately. If you have 
any suggestions, please, comment.


ddalcu:
cool, this is just what I was looking for !

rumbu:
An example view:
Example screenshot

arshaw:
thanks for this. and thanks for the docs (in #2833)

dist/fullcalendar.js
```diff
@@ -5728,6 +5728,12 @@ DayGrid.mixin({
 				else if (typeof clickOption === 'string') { // a view name
 					view.calendar.zoomTo(date, clickOption);
 				}
+			})
+			.each(function() {
```
arshaw: no need to do an `each` for just one element here, which you then have to do `$(this)` for. can you use the `$("<a>")` previously created, directly?

dist/fullcalendar.js
```diff
@@ -5728,6 +5728,12 @@ DayGrid.mixin({
...
+				var renderCallback = view.opt('eventLimitRender');
```
arshaw: could you please use `View::trigger` to call the handler(s)?

arshaw:
I'd prefer an API more similar to [eventLimitClick](https://github.com/rumbu/fullcalendar/blob/b5a860c01a567878f6464a38c92287c9ea5f64c1/src/common/DayGrid.limit.js#L207-L213) where an object with named arguments is passed to the handlers.

arshaw:
Could resubmit this PR without committing the `dist` files? only changes in `src`

rumbu:
Committed the changes based on @arshaw suggestions.

arshaw:
code looks great!

the last piece of the puzzle is adding some simple automated tests to prevent regressions in the future
^
writing tests can sometimes be confusing, so lemme know if you need help

arshaw:
we now have these hooks in v5: https://fullcalendar.io/docs/more-link-render-hooks

PR #2834, opened 2015-08-20
eventLimitRender(callback)

rumbu:
Didn't compiled this. What tool has been used for that?


arshaw:
it's a custom compilation tool i only have on my own machine :( i plan to move away from it

closing this for organizational reasons, but will use it when i tackle #2833

Issue #260, opened 2015-08-05
dayClick doesn't work when using fastclick (github.com/ftlabs/fastclick)


invalidusernam3:
The dayClick event in the fullCalendar configuration stops working when using fastclick (github.com/ftlabs/fastclick) on iOS safari. The next/previous month clicks work fine, but the dayClick event stops working

mlumbroso:
+1 was able to reproduce too.

mlumbroso:
For the record, was having the same problem with a datepicker, here is the solution they came with : https://github.com/dbushell/Pikaday/issues/101

Any chance this is similar? I will try to look in the source and tell if that is the case.

mlumbroso:
So a quick look told me this is definitely the same issue : 
Here is the pull request : #2839

arshaw:
i am nearly certain this is fixed since v2.7.0 introduced official touch support. can someone verify? i will repoen this ticket if issues still exist

PR #2839, opened 2015-08-26
Fix bug with fastclick and dayclick on iOS #260

mlumbroso:
Dayclicks not working when using Fastclick, I fixed it the same way as Pikaday did : https://github.com/dbushell/Pikaday/issues/101

### DIFF
--- a/dist/fullcalendar.js
+++ b/dist/fullcalendar.js
@@ -3296,6 +3296,16 @@ var Grid = fc.Grid = RowRenderer.extend({
 		// attach a handler to the grid's root element.
 		// jQuery will take care of unregistering them when removeElement gets called.
 		el.on('mousedown', function(ev) {
+			ev.preventDefault();
+			if (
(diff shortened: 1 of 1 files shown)


dist/fullcalendar.js
```diff
@@ -3296,6 +3296,16 @@ var Grid = fc.Grid = RowRenderer.extend({
 		// attach a handler to the grid's root element.
 		// jQuery will take care of unregistering them when removeElement gets called.
 		el.on('mousedown', function(ev) {
```
arshaw: if the handlers is exactly the same, couldn't we do `el.on('mousedown touchstart', ...` ?

arshaw:
would be great if the files in `src/` were modified instead

arshaw:
closing this PR but will reference when addressing #260

gmunumel:
Still with this fix I can't get it work. I'm using iOS9

Issue #2848, opened 2015-09-02
Drop callback don't get the event but a moment object


pascoual:
The event param is not the event but a momentjs object:
```
drop: function(event, delta, revertFunc, jsEvent, ui, view) {
  console.log(event);
}
```

Get a MomentJS object.

arshaw:
this is intended behavior:
http://fullcalendar.io/docs/dropping/drop/

you might be wanting the `eventDrop` callback, which is different

arshaw:
you might be needing `eventReceive` as well

PR #2849, opened 2015-09-02
Fix: #2848 - Drop callback don't get the event but a moment object

pascoual:


arshaw:
intended behavior. use `eventDrop` or `eventReceive`

PR #2855, opened 2015-09-05
Vertical resource view (churchdesk) for v2.3.2

drago1:
hi,
fullcalendar with your resourceView is very useful. do you have plans to port resourceView to newer fullalendar's version? Maybe 2.3.2


espen:
Given that Fullcalendar Scheduler introduces resources it should follow that syntax.

I also think that this should be made a custom view and wrapped as a plugin. http://fullcalendar.io/docs/views/Custom_Views/

Lots of good work is done here, would love to see this being available as an easy to implement plugin to FC.

arshaw:
closing this issue for the reasons stated here: https://github.com/fullcalendar/fullcalendar/pull/195#issuecomment-223777717

Issue #2878, opened 2015-09-24
IE9 issue:draggable highlight background and resizable event label are not properly displayed


likereadboy:
When I test official demo (http://fullcalendar.io/js/fullcalendar-2.4.0/demos/agenda-views.html) by Internet Explorer 9, I found that draggable highlight background spans more than has expected.
![image](https://cloud.githubusercontent.com/assets/11569116/10065935/77dc5002-62bf-11e5-846d-3fb1fb918e01.png)

And the resizable event label was "dance" when I expand it to the last cell in week row.
action before:
![image](https://cloud.githubusercontent.com/assets/11569116/10065950/99ef68a0-62bf-11e5-8d84-ea132400d5a1.png)
action after:
![image](https://cloud.githubusercontent.com/assets/11569116/10065953/a779bfd4-62bf-11e5-9714-a262cb631f74.png)
and action finished:
![image](https://cloud.githubusercontent.com/assets/11569116/10065959/b637fc3e-62bf-11e5-9bd7-6da1dc6f83f4.png)

I also found someone was questioned the similar issue on stackoverflow:
http://stackoverflow.com/questions/26834462/fullcalendar-selectable-highlight-in-ie-9-spans-more-than-a-cell-on-select

arshaw:
hmmm, i was unable to reproduce...
![screen shot 2016-06-04 at 8 16 28 pm](https://cloud.githubusercontent.com/assets/56520/15803416/4d8a3162-2a91-11e6-8244-b64d3f275e0b.png)

is this still a problem for you?

fcharrier:
No it's ok, we stopped IE9 support

PR #2881, opened 2015-09-29
fix IE9 Issue:draggable highlight background and resizable event label are not properly displayed

likereadboy:
Issue #2878


arshaw:
I would really like to avoid setting explicit widths on the cells and rather let the browser take care of it.

see my latest comment here about being unable to reproduce: https://github.com/fullcalendar/fullcalendar/issues/2878#issuecomment-223790737

Issue #2853, opened 2015-09-03
Change how month-view events are positioned. Avoid gaps.


joshualambert:
I'm in the process of implementing an activity calendar for a nursing home that will be printed on sheets of LEGAL sized paper. I'm having a difficult time making everything fit into the calendar, but this is because in the new versions of FullCalendar, it renders events on a row-by-row basis for each week inside of tables instead of the DIV method that worked so much better for events that needed to word-wrap. This new feature is resulting in a ton of whitespace as seen in my image example below:

![2015-09-02_1304](https://cloud.githubusercontent.com/assets/1421842/9662359/ca43dd3e-5224-11e5-88e8-fc710b37346d.png)

Does anybody know if it's possible to force FullCalendar to render like it did back in 2.0.0? I'd like to use the latest version, but this change using tables is terrible for longer events, and may require me to use an older version of the library.

joshualambert:
Here's a JS fiddle example of this problem in action:
http://jsfiddle.net/ho4ccfmv/3/

Notice the 2nd and 3rd events on the 17th have white space causing this cell to be taller than it should.

arshaw:
See the really old changelog entry:
https://github.com/fullcalendar/fullcalendar/blob/master/CHANGELOG.md#v210-2014-08-25

> Events are styled to be one line by default ([1992]). Can be undone through custom CSS, but not recommended (might get gaps [like this][111] in certain situations).

arshaw:
so, in a way, known behavior. but keeping this open to discuss potential solutions/workarounds

arshaw:
but the quick fix it keep the default behavior of having the events be one line

joshualambert:
One thought I had was to why not have each DAY CELL in a row contain a child table, where each row inside the child table could contain an event. I think this is the happy medium that avoids using floats, and leverages tables to solve this problem. Thoughts?

joshualambert:
My current approach is really dirty, but I haven't been able to find another way-- I combine all the events for a day together, and pass only a single event to each day from my API with the API layer adding the timestamps as part of the event title. Not the cleanest thing, but at least I got the results I was after. See here: 

<img width="1426" alt="2015-10-11_0850" src="https://cloud.githubusercontent.com/assets/1421842/10417073/25dc87ca-6ff5-11e5-8cdc-de4a9b1cd3e2.png">

arshaw:
[...]

PR #2917, opened 2015-10-27
ensure header elements are always equally spaced

derekm:
FullCalendar CSS gives all header elements a left margin of 0.75em.

With a header config of:

```
header: {
    left: '',
    center: 'prev title next',
    right: ''
},
```

the `title` is an `<h2>`, giving a calculated font-size of 1.5em, while the `prev` and `next` buttons have a `.fc button` font-size of 1em.

This causes the title to have a larger left margin because its font-size is bigger. Setting header element left margins to 0.75**r**em ensures that all header elements will have equal spacing based on the root document font-size.

### DIFF
--- a/src/main.css
+++ b/src/main.css
@@ -29,7 +29,23 @@
 .fc .fc-toolbar > * > :first-child { /* extra precedence to override button border margins */
 	margin-left: 0;
 }
-	
+
+.fc .fc-toolbar > * > button + h2 {
+	margin: 0 .75em;
+}
+
+.fc .fc-toolbar > * > h2 + button {
+	margin-left: 0;
+}
+
+.fc .fc-toolbar > * > h2:first-child {
+	margin-right: .75em;
+}
+
+.fc .fc-toolbar > * > h2:last-child {
+	margin-right: 0;
+}
+
 /* title text */
 
 .fc-toolbar h2 {
--- a/src/main.css
+++ b/src/main.css
@@ -22,7 +22,7 @@
 /* the things within each left/right/center section */
 .fc .fc-toolbar > * > * { /* extra precedence to override button border margins */
 	float: left;
-	margin-left: .75em;
+	margin-left: .75rem;
 }
 
 /* the first thing within each left/center/right section */
--- a/src/main.css
+++ b/src/main.css
@@ -22,7 +22,7 @@
 /* the things within each left/right/center section */
 .fc .fc-toolbar > * > * { /* extra precedence to override button border margins */
 	float: left;
-	margin-left: .75rem;
+	margin-left: .75em;
 }
 
 /* the first thing within each left/center/right section */


arshaw:
was sort of by design. was hoping there would more margin surrounding the title, because it is bigger.

maybe i'll revive this issue at some point, but unfortunately `rem` units are not supported by IE8, which the project attempts to support.

derekm:
I'll open this as an issue, so an appropriate solution can be discussed, because it is a real issue that imparts poor layout dimensions.

arshaw:
oh i didn't realize it creates uneven space. forgive me for misreading. maybe we shouldn't use `rem` but some other solution should work. maybe solution is to make the margin symmetrical.

arshaw:
note: #2921 contains screenshots and more code snippets

derekm:
I've pushed commits that use CSS 2.1 "adjacent sibling selector" (as opposed to less-correct CSS 3 "general sibling selector" from the example in the linked issue) to achieve symmetry in the padding of header title.

src/main.css
```diff
@@ -29,7 +29,23 @@
...
+
+.fc .fc-toolbar > * > h2:first-child {
+	margin-right: .75em;
+}
+
+.fc .fc-toolbar > * > h2:last-child {
+	margin-right: 0;
+}
+
 /* title text */
```
arshaw: I really like this adjacent sibling technique and would rather use it all-the-way and completely drop the existing [first-child technique](https://github.com/derekm/fullcalendar/blob/03d0886e77c527ef9ccdaa255d80443ce111e54b/src/main.css#L22-L31). I think that could boil down all the spacing CSS to only 3 statements:

```
.fc .fc-toolbar > * > * + * {
    margin-left: .75em; /* space between controls */
}

.fc .fc-toolbar > * > * + h2 {
    /* an h2 has larger font-size, thus larger em units, so make it control the right margin to maintain symmetry */
    margin-right: .75em;
}

.fc .fc-toolbar > * > h2 + * {
    margin-left: 0; /* voided because h2 already provides this spacing */
}
```

Could you please verify and test if you think it's a good idea?

arshaw: i updated my above code a bit

arshaw:
Now that fullcalendar has IE9+ support, rems are now a possibility. Though if the page's font-size is set on something other than the root html element, rems don't adapt. I'd like the spacing to be dependent on the font-size of the *calendar's* element.

Rems don't cut it for that. And I don't want to do the messy adjacent sibling technique I was previously a fan of.

The best solution IMO is better use of containers in combination with flexbox.

I'll likely tackle that when doing a CSS refactor for #2853.

In the meantime, this solution as is stands is not something I'd merge into master, so I'll close. I'll keep the bug ticket #2921 open of course.

PR #2918, opened 2015-10-28
Commonjs branch to locale UMD wrapper

Tarabyte:
An error occurs when requiring separate locales in Node/CommonJS environment (say browserify). The error is caused by UMD wrapper that falls back to globals w/o checking for `exports` to be defined.

Here is a repository demonstrating the issue 
https://github.com/Tarabyte/fullcalendar-locale-fails-in-commonjs-env

and a fiddle http://jsfiddle.net/tarabyte/h9o2unnp/ (open console)

```javascript
else if (typeof exports === 'object') { // Node/CommonJS
    module.exports = factory(require('jquery'), require('moment'));
}
```


Tarabyte:
BTW I've noticed that `grunt languages` generates files like this
```javascript
var ru = moment.defineLocale('ru', {...});

return ru;

//unreachable and will be stripped by uglify
$.fullCalendar.datepickerLang(

$.fullCalendar.lang(
```

Am I missing something?

peternowee:
I found 11 automated karma tests related to non-default languages failing because of `uglify` throwing away the FullCalendar-specific locale settings. I can confirm that this patch fixes that.

arshaw:
closed via https://github.com/fullcalendar/fullcalendar/commit/b7f1fca9c32fa4e1c8d3160e2f83d3fd93ca846a

i took a different approach for the umd issue btw. https://github.com/fullcalendar/fullcalendar/commit/a26e17f1fb309e7cb5d21a336e117c93791e0618

Issue #2926, opened 2015-11-03
drop event returns mouseDown jsEvent


martin-langhoff:
the jsEvent returned is the mouseDown jsEvent. This is not documented, and I expected (and need!) the mouseUp event.

Currently working on a patch for this, adding a second jsEvent in the callback. If there's interest in merging it, I'll be happy to update docs, tests.

martin-langhoff:
@arshaw -- so updated PR #2927 to tip of master (at v2.6.0). Should be straightforward, if you think this is the right fix.

arshaw:
@martin-langhoff i appreciate the effort to make this backwards-compatible by adding an additional parameter, but I feel that the current behavior is a "bug"... any sort of final drop handler should only ever get the mouseup event. should not get the mousedown event at all. this is consistent with how `eventDragStop`, `eventDrop`, and `eventResize` do it.

So, I would simply take the event that `listenToExternalDrag` gets from `dragStop` and pass that into `reportExternalDrop` instead of the current `ev` (the mousedown). Make sense?

martin-langhoff:
+1 from me!

As the project is yours, you get to decide compatibility or fix on subtle API bugs such as these. I took the conservative approach.

arshaw:
thanks for taking the conservative approach. i think we can move ahead with this slightly-breaking approach. if you want to make another PR that threads `dragStop`'s `ev` parameter through, I would accept it.

arshaw:
this is fixed for v4: 8715e098fad09a77c9e0426d4cf348d326ae13bb

i'll comment again on this issue when v4 is officially released

arshaw:
[V4 was officially released a few days ago!](https://fullcalendar.io/blog/2019/03/v4-officially-released)

PR #2927, opened 2015-11-03
triggerExternalDrop() - include the drop event in payload

martin-langhoff:
this adds a 'DropEvent' param to the drop callback, in addition to
the current event param, which contains the DragStart / MouseDown
event.

-- Tentative PR for #2926

### DIFF
--- a/src/common/Grid.events.js
+++ b/src/common/Grid.events.js
@@ -484,9 +484,9 @@ Grid.mixin({
 				enableCursor();
 				_this.unrenderDrag();
 			},
-			dragStop: function() {
+			dragStop: function(dropEvent) {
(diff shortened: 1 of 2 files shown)


martin-langhoff:
@arshaw -- this patch assumes that the current behaviour is reasonable (returning the start event), preserves current API, adds a dropEvent param. It does not break the existing API.

If you think it's a bug that the 'ev' param to the drop callback is the drag event, it's a trivial patch to put the drop event there instead.

~thanks!

martin-langhoff:
@arshaw - updated to 2.6.0 -- merge me! merge me! ;-)

arshaw:
closing this PR because I feel we can take a simpler approach. I commented on the original issue.

PR #2974, opened 2015-12-17
Added context menu (right click) event listener.

wenkepaul:
Added callback for context menu click event.


cojack:
:+1:

arshaw:
thanks for this. sorry for the delay. would you mind adapting this to the latest way of doing things?
https://github.com/fullcalendar/fullcalendar/blob/v2.7.3/src/common/Grid.events.js#L177-L184

also, would really like a way to abstract the binding of these very simple actions. attaching `click` is almost identical to attaching `contextmenu` and i'd like to make it easier for similar events in the future.

PR #2979, opened 2015-12-22
Added Luxembourgish (lb) translations

mweimerskirch:


lang/lb.js
```diff
@@ -0,0 +1,16 @@
+
+$.fullCalendar.lang("lb", {
+	buttonText: {
+		month: "Mount",
+		week: "Woch",
+		day: "Dag",
+		list: "Terminiwwersiicht",
+		prev: "zréck",
```
arshaw: prev/next/today are actually derived from jquery-ui datepicker's translations, so no need to include them here. ref: https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-lb.js#L17-L19

Thusly, could you please remove yours? Also, I hope it's okay that jquery-ui's starts with capital letters, which I like better for consistency.

lang/lb.js
```diff
@@ -0,0 +1,16 @@
...
+		next: "weider",
+		today: "haut"
+	},
+	allDayText: "Ganzen Dag",
+	eventLimitText: function(n) {
+		return "+ " + n + " méi";
```
arshaw: this will display `"+ 5 méi"` for example with a space between the + and the 5. Is this absolutely necessary? If not, you'll be able to simply do `eventLimitText: "méi"`

arshaw:

arshaw:

Issue #2980, opened 2015-12-22
moment.localeData() returns null for locales other than en


peternowee:
Some automated karma tests related to locales are still failing in my setup (even after applying #2918). For example:

    PhantomJS 1.9.8 (Linux 0.0.0) columnFormat when lang is French should have the translated dates FAILED
    	Expected 'dim. 5/11' to be 'dim. 11/5'.
    	Error: Expected 'dim. 5/11' to be 'dim. 11/5'.
    	    at /home/peter/Development/fullcalendar/tests/automated/columnFormat.js:85

I did some manual testing and found `moment.localeData('fr')` returning `null`, even with `dist/lang-all.js` loaded. Loading `lib/moment/locale/fr.js` separately in the HTML (in combination with #2918) solves the problem:

    <script src='fullcalendar/lib/jquery/dist/jquery.js'></script>
    <script src='fullcalendar/lib/moment/moment.js'></script>
    <script src='fullcalendar/lib/moment/locale/fr.js'></script>
    <script src='fullcalendar/dist/fullcalendar.js'></script>
    <script src='fullcalendar/dist/lang-all.js'></script>

I guess there is still some problem with Moment locales and `generateLanguages.js`, but I don't know enough about all this factory wrapping etc. to solve it.

peternowee:
Let me know if you're not seeing that particular test fail, then I will need to check my own setup.

peternowee:
Closed in favor of PR #2983, which fixes this.

PR #2983, opened 2015-12-24
Adapt generateLanguages to changed Moment locales

peternowee:
Since around version 2.10.0, the [structure of Moment.js locale files
has changed][1], assigning the result of `moment.defineLocale()` to an
intermediate variable before returning it.

This commit changes the regular expressions in FullCalendar's
`generateLanguages` script to expect the new format.

In my setup, this patch by itself on current master reduces the number 
of failing tests from 20 to 2.

Closes fullcalendar/fullcalendar#2980

I think this also removes the need for [commit 3fd5318][2] of pull 
request fullcalendar/fullcalendar#2918, but do note that that request 
contains another commit as well.

[1]: https://github.com/moment/moment/commit/446cee8ca024507f06ec8d6f7bef361458c70cae/locale/fr.js#diff-2bc288913ddf5c7036261781142e800a
[2]: https://github.com/Tarabyte/fullcalendar/commit/3fd53185c67f01afd3441ecd4a20bd62df8f5ed1

### DIFF
--- a/build/tasks/generateLanguages.js
+++ b/build/tasks/generateLanguages.js
@@ -133,12 +133,18 @@ module.exports = function(grunt) {
 			}
 		);
 
-		// replace the `return` statement so execution continues
-		// compatible with moment-pre-2.8
+		// remove the var/return wrap
+		var lc;
 		js = js.replace(
-			/^(\s*)return moment\.(defineLocale|lang)\(/m,
-			'$1(moment.defineLocale || moment.lang).call(moment, '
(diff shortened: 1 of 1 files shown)


arshaw:
this is the solution i ended up taking: https://github.com/fullcalendar/fullcalendar/commit/a26e17f1fb309e7cb5d21a336e117c93791e0618

(uses a simply JS closure around the moment JS)

peternowee:
That's much simpler indeed. Didn't know that trick yet :) thanks.

Issue #2941, opened 2015-11-16
Option to display week number in day cell


peternowee:
This is a feature request/discussion starter for an option for FullCalendar (basicDay, basicWeek and month views) to display the week number in an existing first column, such as in the top left of the first day of the week, similar to this image from OS X's calendar app:

![Image from OS X's calendar app, from https://github.com/owncloud/calendar/issues/157#issuecomment-25627615](https://camo.githubusercontent.com/e307344f60d02209e2dd717197c8ad2cd560358f/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f313235303534302f313236323231362f31303033643130302d326333632d313165332d386461632d3134366135383530323665612e706e67)

This feature would benefit users of the downstream ownCloud calendar app. The current display of week numbers in a separate column is not acceptable for @jancborchardt (ownCloud design lead) and @georgehrke (ownCloud software engineer), and therefore any implementation of week numbers in ownCloud is currently blocked. See owncloud/calendar#929.

I suggest a new option `basicViewWeekNrPosition`, with (for now) two possible values:

* `weekNrSeparateColumnLeft` (default): Display week numbers in a separate column on the left. This is the current behavior, and therefore should remain default.
* `weekNrExistingFirstColumn`: Display week numbers in an existing first column. In case of month and basicWeek views, week number(s) will be in the top left of the cell(s) of the first day(s) of the week. In case of basicDay view, the week number will be in the top left of the day shown.

The option should not have any effect on the agendaDay and agendaWeek views, which currently show the week number in the column of the time slots, meaning it is not taking up any extra room.

I am willing to do more, though I must say my time is limited. This is just a ticket for discussing the feature's specs before coding, as requested in the [guidelines for contributing](https://github.com/fullcalendar/fullcalendar/blob/master/CONTRIBUTING.md#contributing-features).

Hope to hear from you soon.

arshaw:
I really appreciate the thorough feature request. I wish there was more time, but this doesn't quite stack up against many of the other bugfixes/feature-requests in terms of priority, so doing work to reach an agreement and implement and eventually merge will likely take a while. If I were you, I'd take full ownership on your end to get it good-enough for your specific project, and then we can take it from there.

peternowee:
Ok, I will give it a try and let you know once I have a working version. Perhaps you can assign this issue to me?

peternowee:
[...]

Issue #2458, opened 2015-08-19
When weekNumberCalculation is ISO, default firstDay to 1


arshaw:
Originally reported on Google Code with ID 2193

```
Correct ISO Calculation: http://jsfiddle.net/x68507/4DvUD/
Wrong ISO Calculation: http://jsfiddle.net/x68507/KmdBF/

When dealing with ISO week calculations (weekNumberCalculation:'ISO') and having a
week start of Sunday (firstDay:0), the week number calculation is off by 1 week.  Technically,
the ISO standard is for a week starting on Monday.  December 28, 2015 should be WEEK
53 and it is correct when "firstDay:1" but is incorrect when "firstDay:0".
```

Reported by `michael.r.humphreys` on 2014-06-26 17:19:13

Imported with **1** stars.

peternowee:
I don't think this is really a bug. You choose ISO 8601 for week number *calculation*, which as you said defines Monday as the first day of the week. However, you then choose to *display* your weeks in rows with columns Sunday through Saturday. As a result, each row will cover two partial ISO weeks and no week number can ever be correct for all days in that row. While you are expecting the week number to be that of the Monday (the second day from the left in your example), others may be expecting the week number to be that of the closest day column (Sunday in your example). This becomes more obvious if you set firstDay to Thursday, for example.

Anyway, I am working on an option to display the week number in the top left of a day cell, and for this use case it would be best if that would be the top left of the first day of the week according to the week number calculation method. In this case, that would be the top left of Monday, even if the displayed week starts on Sunday. See fullcalendar/fullcalendar#2941.

arshaw:
thanks for the explanation of this @peternowee

arshaw:
reviving this issue because i see it has some duplicates, indicating demand i didn't realize
#2168 #2328

arshaw:
i feel the most straight forward solution here is to detect if `weekNumberCalculation` is set to `'ISO'` and if so, default `firstDay` to `1` if it hasn't been explicitly specified.

arshaw:
forgot to close this issue. implemented in https://github.com/fullcalendar/fullcalendar/pull/3255

Issue #424, opened 2015-08-19
Clickable day numbers/headers for easier navigation


arshaw:
Originally reported on Google Code with ID 153
```
I have some ideas on how to improve the navigation possibilities between
the different views. As I am not too much of a JS coder, maybe you might
consider coding them:

- weekView: Make headers with names of days clickable to switch to this
day's view.
- monthView: Add a new column with the week numbers to the left and make
those clickable to switch to this week's view.
- In addition to the arrows to go to the next and previous
year/month/week/day a select box might be helpful. In the month view it
would contain the past X and next Y months to choose from. The same for
weeks with the previous X and next Y weeks.
```

Reported by `renzo.lauper` on 2009-10-29 08:43:04

Imported with **41** stars.

SaeX:
Workarounds/hacks for related matters can be found on http://stackoverflow.com/questions/31151320/fullcalendar-detect-click-in-week-or-day-view. For clickable day numbers, see http://stackoverflow.com/questions/8419116/how-to-make-day-value-clickable-in-month-view / http://jsfiddle.net/cjaarmgp/1/. 

Especially the below requests would be very handy as a native feature.
`    - weekView: Make headers with names of days clickable to switch to this day's view.`
`    - monthView: Add a new column with the week numbers to the left and make those clickable to switch to this week's `

arshaw:
thanks for posting those articles to show what kind of hacks are necessary to get around this issue

arshaw:
implemented in v3! https://fullcalendar.io/blog/2016/09/feature-packed-releases/

[navLinks](https://fullcalendar.io/docs/mouse/navLinks/)

funkytooo:
> ```
> My Dup'd issue (403), was to change monthview where you can click on the day number,
> and it opens up the agendaDay view for that specific day ...
> 
> Here is the code I changed/added to make it happen, in case anyone else wants to add
> it.
> 
> On my calendar page, I used the following:
> 
> function moveToDay(date){
>     var toDate = new Date(date);
>     //alert(toDate);
>     $('#calendar').fullCalendar( 'changeView', 'agendaDay' );
>     $('#calendar').fullCalendar( 'gotoDate', toDate );
> }
> 
> In fullcalendar.js, I modified the following:
> 
> <div class='fc-day-number'>"+aO.getDate()+"</div>
> 
> with:
> 
> <div class='fc-day-number'><a href='#' onClick='moveToDay(\""+aO+"\");return
[...]

PR #3024, opened 2016-02-01
Option to display week number in day cell

peternowee:
As discussed in issue fullcalendar/fullcalendar#2941:

These changes add the option to display week numbers in the day cells instead of along the side. The user option `basicViewWeekNrPosition` is introduced, which can have two possible values:

- `weekNrSeparateColumn` (default): Display week numbers in a separate column on the left. This is the current behavior, and therefore should remain default.

- `weekNrDayCell`: Display week number in a day cell. In case of month and basicWeek views, week numbers will be in the top left of the cells of the first day of the week according to the calculation method. In case of basicDay view, the week number will be in the top left of the day shown, regardless of the day of the week.

General notes:

- The option does not affect the `agendaDay` and `agendaWeek` views, which currently show the week number in the column of the time slots, meaning it is not taking up any extra room.

- With `weekNrDayCell`, the week number is not shown if the first day of the week according to the calculation method is not in view, for example:

  - When showing only Monday through Friday while the week number changes on Sunday.
  - In `basicDay` view, when viewing any other day than the first day of the week.

- For `weekNrDayCell`, the relationship with `weekNumberCalculation` and `firstDay`:

  - With ISO: Week number always shown on Monday, as per ISO 8601. Is not influenced by the `firstDay` option.

  - With custom function: A sequence of 7 days is tested with the custom function to determine on which day of the week the week number changes. Is not influenced by the `firstDay` option.

  - With local: If `firstDay` is set, the week number will be shown on that day. If not, the week number is shown on the first day of the week according to the locale (lang). If that is also not set, it is shown on Sunday.

Notes regarding HTML classes:

- The day numbers and their class `fc-day-number` were moved from their own `td` element to a new `span` within that `td`. The `td` got class `fc-numbercell` instead and can contain `span` elements for `fc-day-number` and `fc-week-number`.

- As a result, two different elements that can now have the class `fc-week-number`, depending on the week number display:

  - `weekNrSeparateColumn` (default): A `td` element with class `fc-week-number`, which contains a `span` element without class. (Same as before.)

  - `weekNrDayCell`: A `span` element with class `fc-week-number` inside a `td` element with class `fc-numbercell`.

Existing CSS selectors were made a bit more specific, hopefully preventing unexpected results.

Tests were extended to cover the new functionality.

All automated tests and style checks show no errors.

These changes are supported by downstream ownCloud developers @jancborchardt and @georgehrke (see owncloud/calendar#929).


peternowee:
Some screenshots:

US English locale, month view
20160317 fc 7ec00e8 pn weeknr-in-daycell 5a4fd11 us english month view


US English locale, basicWeek view
20160317 fc 7ec00e8 pn weeknr-in-daycell 5a4fd11 us english basicweek view

caseyjhol:
It seems to me like it'd make more sense to have the number in the bottom left corner to reduce confusion with the date.

peternowee:
@caseyjhol Thanks for your suggestion, but I would prefer to keep it as is for the following reasons:

1. The week number has a different background color, which should help reduce the likelihood of confusion significantly.

2. I doubt that many people will actually experience confusion

  2.1. This design has been discussed in the ownCloud Calendar app repository since 2013 (https://github.com/owncloudarchive/calendar/issues/157 and https://github.com/owncloudarchive/calendar/issues/929), and in the FullCalendar repository since November 2015 (https://github.com/fullcalendar/fullcalendar/issues/2941). Despite many people looking at it, the issue you bring up has not been brought up before.

  2.2. The layout is similar to that of the Mac OS X Calendar app:
mac os x calendar app

  2.3. The existing week number layout in a separate column allows for the same kind of confusion already, but seemingly this was never considered a big enough problem to necessitate a change:
20160318 fc 7ec00e8 separate column us english month view

3. The new layout is entirely optional. FullCalendar users other than ownCloud need to explicitly choose the new layout if they want it.

4. You can create your own layout later. The introduction of the `basicViewWeekNrPosition` setting makes it possible for anyone to propose additional options for other layouts later without affecting the layout of the existing options. For example, you could propose the addition of a `basicViewWeekNrPosition: 'weekNrDayCellBottomLeft'`.

For now, I would prefer to keep bottom left week numbers out of this pull request.

peternowee:
@caseyjhol 

'5. Bottom left week numbers might lead to confusion of their own, because they might be mistaken for 'tabs' attached to the week below it.

jancborchardt:
Also, having the week numbers on the bottom will result in a very busy layout, and another row blocked in the day box which can’t really be used for events. Not an option.

src/Calendar.js
```diff
@@ -337,6 +337,48 @@ function Calendar_constructor(element, overrides) {
 		}
 	})(options.weekNumberCalculation);
 
+	// Store the first day of the week according to the week number
```
arshaw: FYI, you might have been inspired by `_fullCalendar_weekCalc`, but the reason for attaching to the locale is that it's picked up in [moment-ext.js](https://github.com/peternowee/fullcalendar/blob/f3fb8c7a5d014cd3120166a1af2d33c22fd981b8/src/moment-ext.js#L138-L150) so that the `.week()` method reflects the logic.

That being said, is the point of all this logic to correctly compute the start-of-week? I could have sworn moment-ext already handled this, but if not, i'd rather have this logic in moment-ext rather than fullcalendar-specific code.

arshaw:
I'm trying to think of more elegant option names and values. What do you think of this:

`weekNumberColumn`

potential values:

- `true` - (default) - week number is always guaranteed to have its own column. This is literally true for month/basic, and essentially true for agenda views
- `'auto'` - (your new behavior) - will be "smart" about making a week number column, only doing it if absolutely necessary. otherwise it will put the week number in the day cell.

src/basic/BasicView.js
```diff
@@ -12,7 +12,9 @@ var BasicView = FC.BasicView = View.extend({
 	dayGrid: null, // the main subcomponent that does most of the heavy lifting
 
 	dayNumbersVisible: false, // display day numbers on each day cell?
-	weekNumbersVisible: false, // display week numbers along the side?
+	weekNumbersVisible: false, // display week numbers?
```
arshaw: storing a member variable for the general `weekNumbersVisible` is no longer necessary after you break it into two states.

jancborchardt:
> 'auto' - (your new behavior) - will be "smart" about making a week number column, only doing it if absolutely necessary

By that do you mean it will depend on the viewport width?

src/Calendar.js
peternowee: > is the point of all this logic to correctly compute the start-of-week?

Yes, especially because users can set a custom function for week number calculation. I see no other way of knowing the first the day of the week of that custom function other than looping through it to see when the week number changes.

> FYI, you might have been inspired by `_fullCalendar_weekCalc`, but the reason for attaching to the locale is that it's picked up in moment-ext.js so that the `.week()` method reflects the logic.

I see what you mean. `_fullCalendar_weekCalcFirstDoW` is now only picked up in `common/DayGrid.js` `renderNumberCellHtml()`, so perhaps attaching to the locale is indeed not necessary. I could try to use another variable to store the result in.

> I could have sworn moment-ext already handled this, [...]

Just scrolled through moment-ext.js, but I can't find anything similar.

> [...] but if not, i'd rather have this logic in moment-ext rather than fullcalendar-specific code.

Just thinking, are you sure that would not lead to it being run too often? I mean, compared to `week()`, that needs to be called for each new date to determine the week number of that date, the first day of the week only needs to be calculated once for the entire calendar. I am not really sure how often new moment-ext objects are created.

arshaw: ah now i understand. you are basically reverse-engineering the week number computation to get the start-of-week. and this is a primitive integer that gets saves for later use.

this is cool, but i think we can simplify and trust that the expression `calendar.moment(0).startOf('day').day()` will be good enough to compute this. it doesn't take the weekcalc custom logic into account, but I feel that `firstDay` / [lang's DOW](https://github.com/moment/moment/blob/2.13.0/locale/es.js#L74) being in alignment with a custom weekcalc function should be a precondition that the caller must obey. it will cover 99.99% percent of the cases i feel.

arshaw: instead of that chained expression mentioned above, this would be preferred
```
localeData.firstDayOfWeek()
```
http://momentjs.com/docs/#/i18n/locale-data/

peternowee: That would seem easiest, yes, but I remember foreseeing corner cases pop up because of FullCalendar-specific settings. Not only custom functions (I cannot judge how many people are actually using that, and if they really thought about aligning the `firstDay` or locale to their function), but also combinations ISO or locale-based week number calculations and `firstDay` settings. It has been a while and I need to have a good look at this again.

peternowee: IIRC, one part of the problem is that `firstDay` currently influences both which day should be *displayed* in the left-most column, as well as the moment's dow for week number *calculation*, which are two different things in my opinion.

arshaw: I would rather detect if `weekNumberCalculation` is set to `'ISO'` and if so, default `firstDay` to `1` if it hasn't been explicitly specified. If someone specifies a whacky weekNumberCalculation function, they're on their own to make sure `firstDay` lines up. This feels more elegant to me.

src/basic/BasicView.js
peternowee: Ok, fixed with 44d3281.

peternowee:
I renamed the option names and values (commit ba0d9d6), but I have not made any changes in functionality yet.

> `'auto'` - (your new behavior) - will be "smart" about making a week number column, only doing it if absolutely necessary. otherwise it will put the week number in the day cell.

This is not how I set it up so far. Right now, when using the new option, the week number will be _always_ be put in the day cell.

I can change that, but then I need to know what conditions define "absolutely necessary [to make a week number column]"?

arshaw:
nah by `'auto'` i was implying that it won't create a new column in `agenda` because it's already living above the axis column. nothing to do with viewport.

after revisiting this name scheme i came up with, i'm not in love with it. definitely open to other suggestions

src/Calendar.js
peternowee: > I would rather detect if weekNumberCalculation is set to 'ISO' and if so, default `firstDay` to `1` if it hasn't been explicitly specified.

Actually, considering issue #2458 and the other ones you mention there, I would think there is some demand from people who like to keep their calendar with Sunday _displayed_ in the left-most column, while using ISO week number _calculation_. For example, someone in the US working on an international project. You could default `firstDay` to `1` when `weekNumberCalculation` is set to `ISO`, but when those people then override `firstDay` back to `0`, that should still only change which day is in the left-most column, not the day on which the week number changes, because by ISO defintion that will always be Monday. (Of course, those people will probably also prefer to display the week number in the day cell to avoid the confusion that led to issue #2458.)

> If someone specifies a whacky weekNumberCalculation function, they're on their own to make sure firstDay lines up. This feels more elegant to me.

Ok.

Please check my changes in commit b42e8e7.

By the way, I tried extending the `firstDayOfWeek()` and `localeFirstDayOfWeek()` functions in `moment-ext.js` to return `1` in case of ISO, but the functions simply never got hit. These functions belong to the `localeData` object, so perhaps they are not exported by moment.js and cannot be extended at all, or they should be extended in a different manner.

peternowee:
Regarding the option name scheme:
- The words 'week number' just take up a lot of room. From what I understand, abbreviations in identifiers are normally discouraged, unless the abbreviation is very common. That cannot be said about 'wn' of course. Originally, I abbreviated to 'weekNr'.
- It would be nice to have the option names say something about their scope. I originally included 'basicView' in the option name to make clear that it has no effect on Agenda views.

Perhaps an option-array or object could help reduce the amount of options and the length of their names. Here is an example along with some ideas for future extension:

````
weekNumberOptions: {
    basicViewDisplay: 'column', # 'column' (default) or 'cell' (new behavior)
  # agendaDisplay:              # Option that might be added later
  # calculationMethod: 'local', # The current weekNumberCalculation option
  # calculationFirstDay: 3      # Spun off from the current firstDay option
}
````

Just a thought. For me, the name scheme is not of major importance and I am not sure if I will have the time available to help you implement a major change like the one I am suggesting here.

Anyway, I think I addressed all the remarks you made so far. Let me know if you have any other remarks. Also let me know if you would like me to rebase (and squash a few commits at the same time), and if so if you would like that in a new branch and PR, or just change the history of this one.

**Comment by arshaw** [comment]
@peternowee we are getting close!

In regards to renaming the setting, I'm not crazy about the `Nr` abbreviation, but I DO like your idea about scoping it within basicView. But regardless, as you suggest, let's not spend any more time renaming. I'll take responsibility over this before launching.

I WOULD like to change the classNames a bit. Could we rename to `fc-numbercell` to `fc-day-top`? Also, can we attempt to remove the `fc-weeknr-daycell` className? If someone wants to specifically target your week number within the day cell, a compound selector of `.fc-day-top .fc-week-number` will do. If they want to target the column (previous behavior), then can do `td.fc-week-number`.

Resulting DOM structure:
```
<td class="fc-day-top fc-sun fc-past" data-date="2016-06-19">
	<span class="fc-week-number">26</span>
	<span class="fc-day-number">19</span>
</td>
```

I'm about to make more comments inline...

File: src/common/DayGrid.js [hunk 3]
```diff
@@ -137,17 +137,39 @@ var DayGrid = FC.DayGrid = Grid.extend(DayTableMixin, {
...
+			// no numbers in day cell (week number must be along the side)
 			return '<td/>'; //  will create an empty space above events :(
 		}
 
 		classes = this.getDayClasses(date);
-		classes.unshift('fc-day-number');
+		classes.unshift('fc-numbercell');
+
+		if (this.view.cellWeekNumbersVisible) {
+			if ((date._locale || date._lang)._fullCalendar_weekCalc === 'ISO') {
```
arshaw: I'd really like `weekNumberCalculation: 'ISO'` make the `firstDay` setting default to `1` for the global scope of the calendar, not merely in DayGrid's rendering. Can do that [near this part of the code](https://github.com/fullcalendar/fullcalendar/blob/master/src/Calendar.js#L317-L321). That will cause the Monday to render as the first column, which based on the discussion in #2458, is what most people expect.

src/common/DayGrid.js
```diff
@@ -137,17 +137,39 @@ var DayGrid = FC.DayGrid = Grid.extend(DayTableMixin, {
...
+			}
+			else {
+				weekCalcFirstDoW = (date._locale || date._lang).firstDayOfWeek();
+			}
+		}
 
 		return '' +
 			'<td class="' + classes.join(' ') + '" data-date="' + date.format() + '">' +
-				date.date() +
+				((this.view.cellWeekNumbersVisible && (date.day() === weekCalcFirstDoW)) ?
```
arshaw: if you implement the change a few lines above, you'll be able to do `date.weekday() === 0` here

src/common/DayGrid.js
peternowee: Just to clarify: "Default" implies that the setting can be overridden by the user, for example by setting `firstDay: 0`, but it seems you actually want to completely ignore the user's `firstDay` setting in case of `weekNumberCalculation: 'ISO'`, right?

peternowee:
Ok, changed the classes with commits 116e9e3 and 573023d.

See my question about `firstDay` inside the ['outdated diff'](https://github.com/fullcalendar/fullcalendar/pull/3024#discussion_r68211778) above.

src/common/DayGrid.js
arshaw: Correct. If the user provides a `firstDay` value, it should trump all else.

peternowee: Sorry, still not clear. Are you on IRC? I'm PeterNowee on Freenode.

peternowee: The following is not clear to me:

I asked if the user's `firstDay` setting be ignored in case of `weekNumberCalculation: 'ISO'`. You answer "Correct", but then continue by saying that a user-provided `firstDay` should "trump all else", which to me seems to be the opposite of ignoring `firstDay`.

If you really mean to let `firstDay` trump all else, then please detail how you would see that in combination with ISO. I can understand allowing the user to override which day is _displayed_ in the left-most column (for example starting the calendar with Sunday on the left), but you cannot change the first day for week number _calculation_, because that is Monday by ISO definition. Your existing week number calculation routine already ignores any locale/`firstDay` setting in case of ISO.

src/common/DayGrid.js
peternowee: I think `date.weekday() === 0` will go wrong when using ISO in combination with a `firstDay` other than 1 (Monday), because in the current setup of FullCalendar:
- Regardless of `weekNumberCalculation`, `firstDay` always overrides the moment's `dow` and I assume moment uses that `dow` to determine the `date.weekday()`.
- For week number calculation, `moment-ext.js` `week()` does an intermediate check for `weekCalc === 'ISO'`,  and then calls moment's `isoWeek()` instead of `week()`. In other words, `firstDay`/`dow` is already ignored for that purpose (which I think is correct given the ISO definition).

Given this combination, checking for `date.weekday() === 0` should result in the (`isoWeek()`/Monday-based) week number being displayed in the cell of the day set by `firstDay`/`dow`, which may not be Monday. I think that is not desirable.

I think that when using ISO and showing week numbers in the day cells, that should always be on Monday, regardless of `firstDay`/`dow`, just as `firstDay`/`dow` is currently not allowed to influence the ISO week number calculation. To accomplish this, I created an intermediate check for ISO (when setting `weekCalcFirstDoW`), just as there is already an intermediate check for ISO in the `moment-ext.js` `week()` function.

src/common/DayGrid.js
arshaw: I misread your previous sentence.

If the user provides a `firstDay` value, it should trump all else. If the user does not provide a `firstDay` value, then the default should be used. The default should be `firstDay:1` if `weekNumberCalculation == 'ISO'` and `firstDay:0` otherwise.

src/common/DayGrid.js
arshaw: Thank you for the clear explanation. I now see the dilemma with my approach when `weekNumberCalculation: 'ISO'` and `firstDay: 0` ... the leftmost column would be Sunday, and the week number column would be in that Sunday column, but the Sunday belongs to the previous week according to the ISO calculation 😦 .

Can you keep the code as-is, but add a sentence or three of comments explaining why `.weekday()` was insufficient. I would really appreciate it.

src/common/DayGrid.js
peternowee: Ok, I created a separate pull request #3255 for that, because it has broader implications, while week number display in day cells does not depend on it.

src/common/DayGrid.js
peternowee: Ok, see commit 97dad5f.

arshaw:
looks good! thanks for the hard work.

peternowee:
:D Great, thank you too! No need to rebase/squash the commits?

arshaw:
nah, it's okay

arshaw:
In the upcoming release, I plan to implement [Clickable day numbers/headers for easier navigation](https://github.com/fullcalendar/fullcalendar/issues/424). Since that will be a breaking DOM/className release similar to this one, I plan to group them together. Just FYI in case things are going slower than you expect.

peternowee:
Ok, thanks and good luck.

peternowee:
Btw, the dynamic option setting you added to 2.9.0 works great!

arshaw:
This is now in [v3.0.0-beta](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.0.0-beta)!

Please check it out: http://fullcalendar.io/js/fullcalendar-3.0.0-beta/demos/week-numbers.html

arshaw:
I renamed the setting to `weekNumbersWithinDays: true`

peternowee:
Great news, thanks! I tested the v3-beta branch just now, including the build system and it still works as I would expect.

arshaw:
released in v3: https://fullcalendar.io/blog/2016/09/feature-packed-releases/

thanks for all the hard work and for an awesome feature @peternowee

peternowee:
:) My pleasure, I have learned a lot. Thank you for accepting my change.

Issue #3039, opened 2016-02-09
Any way to fill parent element?


aharpervc:
I'm using the scheduler plugin and trying to get it to fill its parent element, which is absolutely positioned to fill _its_ parent. Currently I'm seeing that if I don't specify `height` or `contentHeight` with a numeric value, it'll use the `aspectRatio` property to come up with a height. Is that correct? Is there any way to say "please fill the height of the parent"?

https://github.com/fullcalendar/fullcalendar/blob/master/src/Calendar.js#L669

arshaw:
please follow #2704

Issue #2704, opened 2015-08-19
Option to set calendar height to height of parent element


arshaw:
Originally reported on Google Code with ID 2439

```
It would be helpful if there was an option to have the calendar fill the height of the
containing div. This is useful if you would like a full screen calendar without scrolling
or if the calendar overflows the div then the div will scroll.

Here is my implementation:

https://github.com/arshaw/fullcalendar/pull/206

Here is another implementation that is implemented in almost the same way:

https://github.com/arshaw/fullcalendar/pull/215
```

Reported by `mmarcucio` on 2015-02-10 16:05:07

Imported with **1** stars.

arshaw:
a PR from @aharpervc addressing this issue: https://github.com/fullcalendar/fullcalendar/pull/3040

arshaw:
err, @aharpervc's above issue doesn't actually address this issue. The most elegant API for this issue is to specify `height:'parent'` as opposed to providing a function

aharpervc:
`height: parent` is a special case implementation of #3040

arshaw:
this is now possible in v5. you can set height to `'100%'`. it accepts any css value

PR #3040, opened 2016-02-09
height/contentHeight accept function for dynamic value

aharpervc:
this makes it a lot easier to size the control to its container. for example, implementations could do `return container_el.clientHeight`

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -667,6 +667,10 @@ function Calendar_constructor(element, overrides) {
 	
 	
 	function _calcSize() { // assumes elementVisible
+		if (typeof options.getContainerHeight === 'function') {
+			suggestedViewHeight = options.getContainerHeight() - (headerElement ? headerElement.outerHeight(true) : 0);
+			return;
+		}
 		if (typeof options.contentHeight === 'number') { // exists and not 'auto'
 			suggestedViewHeight = options.contentHeight;
 		}


aharpervc:
This is basically what I'm doing to solve https://github.com/fullcalendar/fullcalendar/issues/3039

flick36:
+1

arshaw:
i agree that #2704 should still be implemented, but this could be cool as well. I'd prefer to see it as the existing `height` or `contentHeight` option that simply accepts a function

arshaw:
Philosophical question: how is this better than [calling the height setter method](http://fullcalendar.io/docs/display/height/)?

I'd rather keep the setter method as the only way to dynamically set an explicit pixel value, and then introduce a `height: 'parent'` option as described in https://github.com/fullcalendar/fullcalendar/issues/2704

aharpervc:
It's a more elegant way to handle a dynamically changing container.

arshaw:
@aharpervc can you provide hypothetical code samples to show how it's more elegant the existing technique of using the setter methods?

aharpervc:
I'll give you my actual production code I'm running with this patch:

```javascript
this.timeline = $(container_el).fullCalendar(_.extend({
    timezone: "utc",
    editable: false,
    aspectRatio: 1.8,

    getContainerHeight: function() {
        return container_el.clientHeight - 15;
    },
    // etc...
```

There's no extra state to track and no extra events to hook up. Whenever calcSize is called internally, the current container client height is used. Now fullcalendar can vertically fill its container both on init and when the browser is resized. There's a `- 15` there because in my actual app, there's kind of a margin, but I didn't want to invent an extra container div to put fullcalendar inside of just for that.

Now, you could certainly re-implement this concept as a static property. However, you'd need to make assumptions on what is the appropriate container element (probably direct parent) and what is the appropriate margins (probably none). Those assumptions *may or may not be true* for all apps and *isn't relevant to the fullcalendar itself*. That's why it's more elegant to delegate the computation to the caller.

arshaw:
ah ok, this is more clear to me now.

how is the height-update initiated? are you manually calling the [render method](http://fullcalendar.io/docs/display/render/) (which readjusts sizing) or are you merely waiting for the built-in window resize handler?

aharpervc:
I'm not explicitly calling render, but also I do have `handleWindowResize` turned on.

arshaw:
ok, this feature makes more sense to me now that i know you are leveraging window resize handling to adjust the height. thanks.

src/Calendar.js
```diff
@@ -667,6 +667,10 @@ function Calendar_constructor(element, overrides) {
 	
 	
 	function _calcSize() { // assumes elementVisible
+		if (typeof options.getContainerHeight === 'function') {
+			suggestedViewHeight = options.getContainerHeight() - (headerElement ? headerElement.outerHeight(true) : 0);
+			return;
```
arshaw: could you leverage the `else if` chain instead of doing a return?

arshaw:
would you be able to write tests for this feature? mandatory to ensure it doesn't break accidentally in the future.

please let me know if i can give you any help/advice.

aharpervc:
I may be able to revisit this at some point, but I don't currently have your test environment set up, or much capacity to investigate it at the moment.

PR #3077, opened 2016-03-06
Use en dash for ranges

flack:
This is just a tiny nitpick, but the long (em) dash in the title has bothered me for a while now. The recommended dash for time ranges is the (shorter) en dash (see e.g. http://www.fonts.com/content/learning/fyti/glyphs/hyphens-and-dashes), and I think it looks nicer, too. Of course users can override the value in configuration, but en dash is a more sensible default than em dash

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -1,7 +1,7 @@
 
 Calendar.defaults = {
 
-	titleRangeSeparator: ' \u2014 ', // emphasized dash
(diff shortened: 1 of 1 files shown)


arshaw:
i agree with this. thank you

arshaw:

arshaw:
Released in [v2.8.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v2.8.0). thanks @flack

PR #3086, opened 2016-03-10
Meteor package

rgnevashev:


.versions
```diff
@@ -0,0 +1,32 @@
+babel-compiler@5.8.24_1
```
arshaw: could we have fullcalendar's build process handle generating this? similar to how the dist files are generated for a release, but not committed to master.

also, what are all these packages? are they dependencies of fullcalendar? i hope not.

bower.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "fullcalendar",
   "title": "FullCalendar",
-  "version": "0.0.0",
+  "version": "2.6.1",
```
arshaw: this should be taken care of by the build process

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "fullcalendar",
   "title": "FullCalendar",
-  "version": "0.0.0",
```
arshaw: this should be taken care of by the build process (that's currently how the version bump works)

package.js
```diff
@@ -0,0 +1,19 @@
+Package.describe({
```
arshaw: im confused by the this file. i thought Dojo owned package.js no? fwiw, here's moment's:
https://github.com/moment/moment/blob/develop/package.js

arshaw:
could you please give more info on what the Meteor package manager is and why FullCalendar should hop onboard?

rgnevashev:
It is meteor package, for atmospherejs...
I think you could be know about meteor...

guidouil:
now meteor is moving it's package management to npm, maybe you should to...

arshaw:
fullcalendar is already installable as an npm module
https://github.com/fullcalendar/fullcalendar/blob/master/package.json
`npm install fullcalendar`

glad to hear meteor will rely on npm. in that case, i'll close this ticket out

PR #3094, opened 2016-03-14
prevent flicker when refetchEvents is called (fix #2558)

### LINKED ISSUE #2558: no flicker on refretchEvents
Originally reported on Google Code with ID 2293
```
Currently we have available only the refetchEvents, which destroys all the events and
then trigger the ajax that catches all events and renders them all again.

This is not the best way to implement an auto update, because the events are "flickering".

The ideal would be to have an option that only UPDATE events. In other words, performs
the Ajax and, according to the result, renders a new event that did not exist, update
an event already rendered or removes an event that no longer exists.
```

Reported by `thiago@thiagomt.com` on 2014-09-23 21:20:57

Imported with **4** stars.

caseyjhol:
This fixes issue #2558. Adds a 'segsRendered' event listener, to prevent the old segs from being unrendered until the new ones are rendered.

Using the "refetchResources" button, see existing behavior here: http://plnkr.co/edit/mVKWwskWG1qiETSeIecl?p=preview

See new behavior here: http://plnkr.co/edit/HWcu73G41fh5BlnZpDFE?p=preview

### DIFF
--- a/src/common/DayGrid.events.js
+++ b/src/common/DayGrid.events.js
@@ -58,10 +58,15 @@ DayGrid.mixin({
 	unrenderFgSegs: function() {
 		var rowStructs = this.rowStructs || [];
 		var rowStruct;
+		var _this = this;
 
-		while ((rowStruct = rowStructs.pop())) {
-			rowStruct.tbodyEl.remove();
-		}
+		this.el.on('segsRendered', function() {
+			while ((rowStruct = rowStructs.pop())) {
+				rowStruct.tbodyEl.remove();
+			}
+
+			_this.el.off('segsRendered');			
+		});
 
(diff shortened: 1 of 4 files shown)


flick36:
This breaks the 'select' functionality, you can select but it won't unselect when you click for example somewhere else

caseyjhol:
Good catch. I might be missing something, but is seems like simply removing `destroyEvents();` from `refetchEvents()` (and ignoring everything else in my PR) solves this problem also.

According to the comment, it's there "so that events are cleared before user starts waiting for AJAX." But, that's not what we want. Having the events cleared before the AJAX call is what causes the flicker.

caseyjhol:
Closing this in favor of #3123.

Issue #2944, opened 2015-11-17
businessHours, doesn't display on agendaDay on weekends


arshaw:
the whole day should technically be grayed out. inconsistent with all the other views, which grays out whole Saturdays/Sundays.

http://jsbin.com/wegafolayo/1/edit?js,output

bug is happening somewhere near here:
https://github.com/fullcalendar/fullcalendar/blob/master/src/EventManager.js#L632

Kyunghwa-Yoo:
Hi, i found this issue while doing my project using fullcalendar.
And i fixed it last night.
i want contribute this bugfix, but i don't know how to do this because it's my first contribution.

first of all, i forked repository and i edited code.
here is what i edited code
https://github.com/Kyunghwa-Yoo/fullcalendar/blob/master/src/EventManager.js#L874

i checked CONTRIBUTING.md document and successfully  `run grunt check`

next, what can i do? 
i create pull request. but i don't know it is right way to contribute bugfix.
please reply to me.

here is my pull request creation.
https://github.com/fullcalendar/fullcalendar/pull/3096

thank you.

arshaw:
this has been fixed and release in [v2.9.1](https://github.com/fullcalendar/fullcalendar/releases/tag/v2.9.1).

PR #3096, opened 2016-03-17
fix businessHours agendaDay

Kyunghwa-Yoo:
hi, i found bug below and i fixed it.
https://github.com/fullcalendar/fullcalendar/issues/2944

and i want to contribute this bugfix but i don't know how to do this because it's my first contribution.

first of all, fork repository.
second, edit src/EventManager.js
third, successfully `grunt check`

and, i think 'create pull request' is next process.
so i create pull request.

here is my edited code
https://github.com/Kyunghwa-Yoo/fullcalendar/blob/master/src/EventManager.js#L874

if it is wrong way to contribute bugfix, then please say to me.

thank you.

### DIFF
--- a/src/EventManager.js
+++ b/src/EventManager.js
@@ -870,6 +870,26 @@ function EventManager(options) { // assumed to be a calendar
 				eventInput.start = null;
 				eventInput.end = null;
 			}
+			else{
+				if(view.name=="agendaDay"){		// check view_name.
+					var dowHash = view.start.stripTime().day();	// get day of week value.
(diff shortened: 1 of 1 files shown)


arshaw:
thank you for this PR. i'll review this PR soon

arshaw:
thanks for this fix. I bet it works, but i'm hoping for a more general solution that doesn't rely on the current visible view being a single day. for example, the bug is present for some people even when there are multiple days being displayed (all of which are outside of businesshours).

does that make sense?

arshaw:
this functionality has been released in v2.9.1

PR #3115, opened 2016-04-12
update Japanese Translation

ounziw:


arshaw:
could you explain what is wrong with the existing translations for Japanese?

ounziw:
The existing translations lack month and week translations.
January, February, ... December / Sunday, ... Saturday will be displayed in English, not in Japanese.

lang/ja.js
```diff
@@ -1,13 +1,18 @@
...
-		month: "月",
-		week: "週",
-		day: "日",
-		list: "予定リスト"
-	},
-	allDayText: "終日",
-	eventLimitText: function(n) {
-		return "他 " + n + " 件";
-	}
+jQuery.fullCalendar.lang("ja", {
```
arshaw: could you use `$.fullCalendar.lang` just like before?

could you also maintain the same type of indentation.. tabs? will make for a cleaner diff

arshaw:
@ounziw i believe you are using fullcalendar's translation system incorrectly. you don't include the dev repo's translation files; you need to include the `dist` translation files included in the ZIP or the bower/npm package. these translation files have the momentjs translations baked in. like this: https://github.com/fullcalendar/fullcalendar/tree/v2.7.3/dist/lang

PR #3123, opened 2016-04-18
prevent flicker when refetchEvents is called (fix #2558)

### LINKED ISSUE #2558: no flicker on refretchEvents

caseyjhol:
This removes the call to the destroyEvents function in refetchEvents(), as discussed here:
- https://github.com/fullcalendar/fullcalendar/issues/2558#issuecomment-132574399
- https://github.com/fullcalendar/fullcalendar/issues/2558#issuecomment-132574410

As the destroyEvents function isn't being used anywhere else, this pull request also removes that.

According to the comment, it's there "so that events are cleared before user starts waiting for AJAX." But, that's not what we want. Having the events cleared before the AJAX call is what causes the flicker. This supersedes #3094.

Using the "refetchResources" button, see existing behavior here:
http://plnkr.co/edit/mVKWwskWG1qiETSeIecl?p=preview

See new behavior here:
http://plnkr.co/edit/c8hqZePSz8KFphI2boAS?p=preview

### DIFF
--- a/tests/automated/refetchEvents.js
+++ b/tests/automated/refetchEvents.js
@@ -1,4 +1,4 @@
-ddescribe('refetchEvents', function() {
+describe('refetchEvents', function() {
 	describe('when agenda events are rerendered', function() {
 		beforeEach(function() {
 			affix('#cal');
--- a/tests/automated/refetchEvents.js
+++ b/tests/automated/refetchEvents.js
@@ -50,6 +50,7 @@ ddescribe('refetchEvents', function() {
 
 		beforeEach(function() {
 			affix('#cal');
+			fetchCount = 0;
 			currentCalendar = $('#cal');
 			options = {
 				now: '2015-08-07',
@@ -76,7 +77,6 @@ ddescribe('refetchEvents', function() {
 
 		describe('and all events are fetched synchronously', function() {
 			it('all events are immediately updated', function(done) {
-				fetchCount = 0;
 				currentCalendar.fullCalendar(options);
 				fetchCount++;
 				currentCalendar.fullCalendar('refetchEvents');
@@ -87,7 +87,6 @@ ddescribe('refetchEvents', function() {
 
 		describe('and one event source is asynchronous', function() {
 			it('original events remain on the calendar until all events have been refetched', function(done) {
-				fetchCount = 0;
 				// set a 100ms timeout on this event source
 				options.eventSources[0].events = function(start, end, timezone, callback) {
 					var events = [
--- a/tests/automated/refetchEvents.js
+++ b/tests/automated/refetchEvents.js
@@ -46,12 +46,12 @@ ddescribe('refetchEvents', function() {
 	describe('when there are multiple event sources', function() {
 		var options;
 		var fetchCount;
-		var currentCalendar;
+		var calendarEl;
 
 		beforeEach(function() {
 			affix('#cal');
 			fetchCount = 0;
-			currentCalendar = $('#cal');
+			calendarEl = $('#cal');
 			options = {
 				now: '2015-08-07',
 				defaultView: 'agendaWeek',
@@ -77,9 +77,9 @@ ddescribe('refetchEvents', function() {
 
 		describe('and all events are fetched synchronously', function() {
 			it('all events are immediately updated', function(done) {
-				currentCalendar.fullCalendar(options);
+				calendarEl.fullCalendar(options);
 				fetchCount++;
-				currentCalendar.fullCalendar('refetchEvents');
+				calendarEl.fullCalendar('refetchEvents');
 				checkAllEvents();
 				done();
 			});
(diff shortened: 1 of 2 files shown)


RinkAttendant6:
:+1: +1, I am unable to upgrade to FullCalendar 2.x because of this issue

thestumonkey:
+1 to merge this PR

arshaw:
this looks great. thanks for this elegant fix.

i'm hoping to get some regression tests in here before merging. can only think of one: when `refetchEvents` is called with one event source that fetches event asynchronously (doesn't return immediately and takes 100ms or so to return), then immediately after the call, the event elements *are still present* (were not cleared yet)

arshaw:
some resources:

lemme know if you have questions/problems if/when you get around to tests

caseyjhol:
I'm adding a test when there are multiple event sources that fetch events synchronously to compare events on the calendar before and after refetchEvents is called to ensure events are properly updated.

I was wondering if you could elaborate on your suggestion for a test up above and what you're expecting with that. Obviously there would be multiple event sources, and one would be asynchronous (e.g. 100ms timeout), but I'm not sure what you want the test to check for specifically. Do you just want to check that 0ms after calling refetchEvents the original events are still present, and 100ms after calling it the updated events have replaced the original events?

arshaw:
@caseyjhol , exactly.

caseyjhol:
@arshaw Tests added. Let me know if you had something different in mind or want me to make any changes.

tests/automated/refetchEvents.js
```diff
@@ -1,45 +1,135 @@
...
+							done();
+						}, 100);
+					}
+				};
+
+				currentCalendar.fullCalendar(options);
+			});
+		});
+
+		function createEvents(id, delay) {
```
arshaw: could you rename this function to indicate that is doesn't return events, but returns a *function* that returns events. maybe `createEventGenerator`

arshaw: the `id` param isn't being used for anything. are you mixing it up with `count` ?

tests/automated/refetchEvents.js
```diff
@@ -1,45 +1,135 @@
...
+					}
 				}
-			}
+			});
 		});
 	});
 
+	describe('when there are multiple event sources', function() {
+		var options;
+		var count;
```
arshaw: could you rename this to `fetchCount` to further indicate what it's used for?

tests/automated/refetchEvents.js
```diff
@@ -1,45 +1,135 @@
...
+		var currentCalendar;
```
arshaw: could you rename this ti `calendarEl`. otherwise, i'm afraid ppl might thing it is a calendar instance object, not a jQuery element

tests/automated/refetchEvents.js
```diff
@@ -1,45 +1,135 @@
...
+
+		beforeEach(function() {
+			affix('#cal');
+			currentCalendar = $('#cal');
+			options = {
```
arshaw: could you initialize `fetchCount` to `0` within here? that seems to be the pattern

tests/automated/refetchEvents.js
```diff
@@ -1,45 +1,135 @@
...
+					}, delay);
+				}
+				else {
+					callback(events);
+				}
+			};
+		}
+
+		// Checks to make sure all refetched events have been rendered
+		function checkAllEvents() {
```
arshaw: instead of making this a separate function, could you *inline* this where it's currently being called? I'd rather have this be less DRY and more literally explain what's going on.

arshaw:
the tests are off to a good start @caseyjhol ! i made some comments

caseyjhol:
@arshaw Updates made. Let me know if there's anything else you notice.

tests/automated/refetchEvents.js
arshaw: things look great. just this one thing here ^

caseyjhol: Hrmmm...weird. If you look at the actual code, that's been updated. For some reason in the "Conversation" section, GitHub didn't update the code.

arshaw:
[Code complete](https://github.com/fullcalendar/fullcalendar/commits/v2.8.0-dev) and [docs written](https://github.com/fullcalendar/fullcalendar/commits/docs-2.0) but not yet released. Will update this thread when released.

arshaw:

thank you @caseyjhol !

Issue #3035, opened 2016-02-08
rename eventLimit to maxEventRows. achieve true max events?


jasonEXA:
Issue created with reference to the 2nd part of [#128](https://github.com/fullcalendar/fullcalendar-scheduler/issues/128).

One other UI issue I encountered (demonstrated in above jsBin link) is when there are multiple events, one of which extends over multiple days and the other is present for a single day for the same resource. The one that extends over multiple days, overlaps the one on the single day. There is no visual indication to the user if there is an event present on that single day.

jsBin output link - http://output.jsbin.com/dunojoqome/

E.g
![image](https://cloud.githubusercontent.com/assets/9958408/12879104/9218ece0-ce54-11e5-9134-e13d137a2989.png)

arshaw:
this looks like some sort of malfunction of the `eventLimit` setting, which you have turned on. does setting it to `false` give you the desired behavior?

arshaw:
you actually have `eventLimit: 3` on basicWeek view, so it's not a malfunction. The number `3` means 3 lines of content, and if there are more that 3 events, then the first two are displayed and the rest are put into the "more" link

arshaw:
I propose the `eventLimit` setting be renamed to `maxEventRows`

p-ob:
Hmm yes that does sound like the same issue (although multi-day events are not required). So to always show just x events per day with the additional events hidden is not the intended purpose? It is always "try to show x, if there are more than x show x-1 as the +n more counts as an event row"?  And if this is the case, the desired effect is a piece of functionality that does not exist for fullcalendar?

arshaw:
@p-ob, you accurately describe the current behavior.

i'll proposed that the current option be renamed to `maxEventRows` and your desired new behavior be called `dailyEventLimit` or something

imransmy:
Hello, I'm using fullcalender in my application. I have created customweek and timelineday as per my client's request. Everything is fine, but i have more than 20 events per day may be it will be more over 100 or 200. In month view im able to view that much of events using view more. In day and week view im not able to add view more. I have added "eventLimit" also tried "maxEventRows " as well as "dailyEventLimit" too. But in week and timeline view it was showing all the events. Please help me.

arshaw:
@imransmy , what you are talking about is a different feature than this ticket. essentially you want eventlimit for the time slots in agenda views, which is currently not possible, nor do I understand how the UI would look for this. could you open a new ticket and explain more?

[...]

PR #3129, opened 2016-04-20
eventLimit more link placement

nickyoung87:
When using the `eventLimit` option we noticed that the "+ X more" text was actually replacing the bottom event rather than being output after the limit was reached. Here is an example using `eventLimit: 3`:

eventlimit

You can see on Thursday that only 2 events are shown. We would expect this to be 3 since that is what the limit is set at.

This PR tries to correct this. I have changed the element to be a `<div>` instead of a `<td>` and then `append()` it to the last element instead of replacing it. I also had to modify the output of the text to be the length minus 1 so it would show the correct number for the "+ x more" text.

Would really love to see if this can be added in. I am open to discussion.

Thanks!

### DIFF
--- a/src/common/DayGrid.limit.js
+++ b/src/common/DayGrid.limit.js
@@ -93,7 +93,7 @@ DayGrid.mixin({
 		var td, rowspan;
 		var segMoreNodes; // array of "more" <td> cells that will stand-in for the current seg's cell
 		var j;
-		var moreTd, moreWrap, moreLink;
+		var moreDiv, moreWrap, moreLink;
 
 		// Iterates through empty level cells and places "more" links inside if need be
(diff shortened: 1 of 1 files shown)


arshaw:
I @nickyoung87 thanks for the PR. Please see the discussion here about renaming this setting to `maxEventRows`: https://github.com/fullcalendar/fullcalendar/issues/3035

I'd rather rename the setting be more in-line with the current behavior rather than display fewer events in cells that might actually accommodate more events due to the extra height adjacent +more links. if that makes sense

arshaw:
in v5 we now have `dayMaxEvents` and `dayMaxEventRows`
https://fullcalendar.io/docs/event-popover

PR #3133, opened 2016-04-25
Bad, bad language. Firewalls hate it!

rahulrulez:
Just a small pull request.. Sonicwall and other **strict** firewalls do not like it causing a lot of broken webapps. I love such comments, firewalls, unfortunately not.


arshaw:
released via 2864289061ca5ed761695aaf8d1132bb8dfd19a5, thanks!

PR #3139, opened 2016-04-27
Make timezone test expect Z suffix for UTC times

peternowee:
Moment 2.13.0 now formats moments in UTC mode with a "Z" suffix instead of +00:00. See https://github.com/moment/moment/pull/3098.

This commit changes the FullCalendar automated test for timezones to expect that new formatting.


arshaw:
closed via 7a09fc019b6fa831002e7aa6db1833675388d3d0, thanks!

PR #3143, opened 2016-04-29
Update EventManager.js

Restartz:
allDay events that have the same start and end date get their end date set to null.
Calling updateEvent on an event will trigger diffDates to run.
$('#calendar').fullCalendar('updateEvent', event, true);
But since the end date is null it will throw a type error since clone can't be called on a null.
The end date will normally (in good use) only be set to null when the start and end date are the same, so the start date (which is always set) can be used here instead of the empty end date.


arshaw:
could you post a JSBin that demonstrates the problem? http://fullcalendar.io/wiki/Reporting-Bugs/

PR #3145, opened 2016-04-29
resolve dependency conflict

loganfreeman:
I tried to run `grunt`. However it failed due to dependency conflict with vow. So I took to resolve it.


arshaw:
fixed by 6143c35fd0f4f6fcd0c5c0c923a038016fa07f13 and other commits in v2.7.1, just released

Issue #3157, opened 2016-05-10
Events are not properly positioned / sized when height is fixed


denniskoster:
Hi,

When I have a fullCalendar instance with a fixed height and events loaded through the `events` option, the events aren't positioned / sized properly. 

Whenever I trigger a window resize, or add / edit an event, all events are repositioned / resized to what they should look like. I have created a js bin here: [http://jsbin.com/sihabusagu/edit?html,css,js,output](http://jsbin.com/sihabusagu/edit?html,css,js,output).

Cheers

nickyoung87:
We just ran into a problem with the height option as well.

When we set the option to `height: 'auto'` the scroll bars will still show up. The same thing happens if we set a pixel value. If we set something like `height: 2000` it will stretch the calendar out but there will still be a scroll bar.

I have moved back to 2.6.1 and this is not an issue there. Tested in 2.7.0 and 2.7.1 and it is broken in both of those releases. 

Not sure if this is completely related to @denniskoster's issue but it seems to be similar enough to post here.

arshaw:
could you post a screenshot of what things look like in the badly-positioned state? could have to do w/ screen/browser size. things look fine on my end

denniskoster:
Hi Arshaw,

Thanks for the reply. I have attached twee screenshots. One with the events incorrectly positioned / sized and one after window resize, where everything is fine.

![image](https://cloud.githubusercontent.com/assets/8804219/15314165/a1f9ccd8-1c10-11e6-8c88-1b86a55403b8.png)
![image](https://cloud.githubusercontent.com/assets/8804219/15314205/be21a368-1c10-11e6-81dd-4c2db04586fa.png)

Thanks in advance!

Dennis

caseyjhol:
I can confirm that I'm getting the same results as @denniskoster (incorrectly positioned events) in both Chrome 50 and Firefox 46 on Windows 10.

denniskoster:
@caseyjhol Thanks for the reply. You are correct, when removing that css events are positioned properly along the grid. However, there's not much point in making the calendar higher if the grid stays that height.

nickyoung87:
Is what I pointed here unrelated to this issue? It seemed like it was at first but now I am getting the impression that maybe it isn't. Let me know if the PR helps with anything here, otherwise I can open a separate issue for what we experienced.

denniskoster:
@nickyoung87 I can't really tell if the two issues are related to be honest. Could you post a jsbin?

[...]

PR #3159, opened 2016-05-11
Add CSS back so height option works as before

nickyoung87:
See also https://github.com/fullcalendar/fullcalendar/issues/3157


arshaw:
could you please describe how things behaved previously, and how they behave now? descriptions of the visual differences along with a [Reduced Test Case](http://fullcalendar.io/wiki/Reduced-Test-Cases/)(s) would be greatly appreciated.

PR #3167, opened 2016-05-16
(feat.) Multiple Business Hours

daniel-nagy:
Allows passing an array to the business hours property.

**Motivation**

* It is not uncommon for a business to have different hours on different days.
* Change business hours when switching days in `agendaDay` view.


daniel-nagy:
My apologies, I made this PR from an older version.

PR #3168, opened 2016-05-16
(feat.) Multiple Business Hours

daniel-nagy:
Allows passing an array to the business hours property.

**Motivation**

* It is not uncommon for a business to have different hours on different days.
* Change business hours when switching days in `agendaDay` view.

I won't be offended if you trash this PR and write your own code to do this, I'm just hoping this feature makes it into master.


daniel-nagy:
Closing because there is much more work to be done to get this to work correctly

PR #3199, opened 2016-06-04
added changeView callback function

github-expert:
Added changeView callback function which will be called every time the user switches view.
The option provided will be like this

```
$('#calendar').fullCalendar({
    ...
    changeView: function (changedView) {
        // do anything with the view that will be changed
    }
    ...
});
```

feature request ticket: #3521

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -582,8 +582,7 @@ function Calendar_constructor(element, overrides) {
 			freezeContentHeight(); // prevent a scroll jump when view element is removed
 			currentView.removeElement();
 			currentView = t.view = null;
-			if(typeof t.options.changeView == 'function')
-				t.options.changeView(viewType);
+			t.trigger('changeView', null, viewType, currentView.type);
 		}
 
 		// if viewType changed, or the view was never created, create a fresh view
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -582,6 +582,8 @@ function Calendar_constructor(element, overrides) {
 			freezeContentHeight(); // prevent a scroll jump when view element is removed
 			currentView.removeElement();
 			currentView = t.view = null;
+			if(typeof t.options.changeView == 'function')
+				t.options.changeView(viewType);
 		}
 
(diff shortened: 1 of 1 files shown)


src/Calendar.js
```diff
@@ -582,6 +582,8 @@ function Calendar_constructor(element, overrides) {
 			freezeContentHeight(); // prevent a scroll jump when view element is removed
 			currentView.removeElement();
 			currentView = t.view = null;
+			if(typeof t.options.changeView == 'function')
```
arshaw: could you please achieve this via `Calendar::trigger` ?

arshaw:
instead of receiving a single string argument of the view type, could this handler receive:

```
changeView: function(newView, oldView) {
}
```

where both arguments are the View objects?

src/Calendar.js
github-expert: like this?
```
if(typeof t.options.changeView == 'function')
    t.trigger('changeView', viewType);
```

arshaw: you don't need the if statement. the `trigger` method will check to make sure the handler exists first before calling it.

correct signature would be:
```
t.trigger('changeView', null, newView, oldView);
```

(we pass in `null` as the "this value", because we don't need one. will populate the calling function's `this` value with the calendar instance which is what we want)

arshaw: see a line like this for an example: https://github.com/github-expert/fullcalendar/blob/5ff8129ccc08d9f170aab1edc6b09ed1d2dbe6a9/src/Calendar.js#L222

github-expert:
i used the trigger function as you suggest in a new commit.
Please check it now.

github-expert:
both arguments are not view objects but the views like `month`, `week`, etc...

arshaw:
awesome.

the final step (and often the most annoying 😄 ) is creating tests:

is this something you think you could take charge of?

github-expert:
can you do that for me?

arshaw:
we now have the `datesSet` callback, which is pretty similar to what this PR requests: https://fullcalendar.io/docs/datesSet

PR #3201, opened 2016-06-05
Added render events for Popover

danielemams:
Call "eventAfterAllPopoverRender" when all events in a Popover are rendered.
Moreover call "eventAfterRender" for every events in Popover.

### DIFF
--- a/src/common/View.js
+++ b/src/common/View.js
@@ -643,11 +643,15 @@ var View = FC.View = Class.extend(EmitterMixin, ListenerMixin, {
 
 
 	// Signals that all events have been rendered
-	triggerEventRender: function() {
+	triggerEventRender: function(inPopover) {
 		this.renderedEventSegEach(function(seg) {
 			this.trigger('eventAfterRender', seg.event, seg.event, seg.el);
 		});
-		this.trigger('eventAfterAllRender');
+		if(!inPopover) {
+			this.trigger('eventAfterAllRender');
+		} else {
+			this.trigger('eventAfterAllPopoverRender');
+		}
 	},
 
 
--- a/src/common/DayGrid.limit.js
+++ b/src/common/DayGrid.limit.js
@@ -264,6 +264,7 @@ DayGrid.mixin({
 
 		this.segPopover = new Popover(options);
 		this.segPopover.show();
+		view.triggerEventRender(true);
 	},
 
 


src/common/View.js
```diff
@@ -643,11 +643,15 @@ var View = FC.View = Class.extend(EmitterMixin, ListenerMixin, {
 
 
 	// Signals that all events have been rendered
-	triggerEventRender: function() {
+	triggerEventRender: function(inPopover) {
 		this.renderedEventSegEach(function(seg) {
```
arshaw: i believe this will trigger the `eventAfterRender` for ALL events, not just those newly displayed in the popover. could you verify?

danielemams: yes. this is called for ALL events.
Now also after every single events in popover is rendered.

arshaw: i dont think most people will want that. the view's initial events will have already been passed thru the `eventAfterRender` handlers. can we change the code to ONLY do it for the newly rendered events in the popover?

danielemams: i dont understand. you said to me to use `eventAfterRender` also for events in the popover. so i have done this pull request. do you want another handler only for the events in the popover? something like `eventAfterRenderPopover` ? so events in the initial view will call `eventAfterRender` and events in the popover will call `eventAfterRenderPopover` ?

arshaw: I apologize if there has been any communication issues.

I want each event to eventually be passed through `eventAfterRender`, but only immediately after it has been rendered. They way you had the code set up, events rendered outside of the popover would be be passed into `eventAfterRender` *again* after the popover was opened.

### example of what I want

- 5 events rendered when the view is initially opened
- those 5 events should be run though `eventAfterRender`. this is already the case
- user opens a popover with 2 events in it
- those 2 events are run through `eventAfterRender`.

### example of what I DON'T want

- 5 events rendered when the view is initially opened
- those 5 events should be run though `eventAfterRender`. this is already the case
- user opens a popover with 2 events in it
- all 7 events are run through `eventAfterRender (**BAD**)

arshaw: does that make sense?

danielemams:
Sorry for my delayed answer.
With my code, you can fire an event for every lines in the popover, and one at the end of the render of all lines in the popover.

This is an example:

http://jsbin.com/heyafuquta/edit?html,js,output

If you open the Console, you can see:

"eventAfterAllRenderPopover"

after the popover is opened, and it renders all the lines.

If you mousover the elems in the popover, you can see the:

eventAfterRenderPopover

that attach the 'title' to the element.

Thanks

arshaw:
as of v4, i can confirm that `eventRender` and `eventPositioned` are both called for each popover event: https://codepen.io/arshaw/pen/eYOzPLK?&editable=true&editors=001

`eventAfterAllRender` has been removed in v4.

PR #3202, opened 2016-06-06
height:'auto' always shows scrollbars in month view on FF

developerbmw:
Currently the scrollbar is always shown in Firefox with a basic calendar like:

```
$('#calendar').fullCalendar({
	height: 'auto'
});
```

This CSS change fixes the issue.

### DIFF
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -359,6 +359,7 @@ hr.fc-divider {
 
 .fc-bg table {
 	height: 100%; /* strech bg to bottom edge */
+	box-sizing: border-box; /* fix scrollbar issue in firefox */
 }
 
 


arshaw:
Would you be able to post a [runnable, stripped-down demonstration of the bug](http://fullcalendar.io/wiki/Reduced-Test-Cases/) you are trying to solve? Would really appreciate it because the time saved reproducing will be time spent fixing. Temporarily closing this ticket until that happens.

FWIW, I'm not able to recreate this on FF on Mac

developerbmw:
@arshaw Here is a demonstration: https://jsfiddle.net/116fh9h0/

The issue appears for me on Firefox 46.0.1 (32-bit) on Windows 10.

https://i.imgsafe.org/66b48aae07.png

arshaw:
thank you. i can now reproduce.

developerbmw:
Thanks for re-opening. Note that this is a PR not an issue - you could make it into an issue if you don't want to merge this PR

arshaw:
thanks for the heads up. looks good to do. i want to do a bit of playing around with this before launch. ill let u know if i find anything

developerbmw:
Thanks. Yeah would be good to do a bit of playing around - I think it may be a bug with Firefox actually.

peternowee:
The fix works when `height = 'auto'`, but I'm still seeing the empty scrollbar when I don't set `height` at all. Using Firefox 45.2.0 (64-bit) on Debian Linux.

peternowee:
Sorry, you can ignore my last comment. After downgrading jQuery from 3.0.0 to 2.1.3 (#3219), the fix also works when `height` is not set.

arshaw:

arshaw:
Released in [v2.8.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v2.8.0). thanks @developerbmw

PR #3211, opened 2016-06-09
Remove underscore (unused)

avindra:
This drops the `underscore` dev dependency. It was being used in the Gruntfile at some point, but it seems like that's history now.


arshaw:
thanks.

arshaw:

PR #3197, opened 2016-06-03
jQuery 3.0 compatibility for removeAttr

avindra:
`removeAttr` will no longer set props to false. [Source](https://jquery.com/upgrade-guide/3.0/#breaking-change-removeattr-no-longer-sets-properties-to-false).

Further, by using `.prop`, it opens up the ability of reducing duplicated logic in `disableButton` and `enableButton`.

Issue #3124, opened 2016-04-18
jQuery 3 support


JoshWillik:
The error message is `TypeError: rangeStart is null`.

**2.3.1**

[demo](http://jsbin.com/nisujeceyo/1/edit)

**2.3.2**

[demo](http://jsbin.com/laqesemewe/1/edit)

JoshWillik:
The bug exists up to 2.6 as well

arshaw:
You are using jQuery 3 beta.

I've update your fiddle to use latest FC and jQuery 2.2.4
http://jsbin.com/wifonupaba/1/edit?html,output
Works as expected.

Will revisit this when jQuery 3 comes out

avindra:
@arshaw : I am also getting this on the jQuery 3.0 Release Candidate 1 (not beta).

avindra:
Still encountering this in the final release of jQuery 3.0. I tried debugging a bit, and it seems to be related to the fact that `currentView.start` is null.

ziemkowski:
This also prevents the calendar from loading for us.

Note that it happens even with [jQuery Migrate 3.0.0](http://jquery.com/download/#jquery-migrate-plugin), which is supposed to make jQuery backwards compatible.

avindra:
@ziemkowski : I noticed that a hacky workaround is to click previous, then click next, which causes our events to render as expected. 

I may end up adding code to simulate the two clicks, because we are shipping soon and want to ship with jQuery 3.0.

avindra:
This is the workaround I'll be using:

```js
// define this global somewhere that it would be in scope:

let stage = 0;
....
// add as a prop to the FC object

			viewRender(view, element) {
				// initial render is fixed now
				if(++stage == 3) return;
				const fc = element.parent().parent();
				const direction = stage == 1 ? 'prev' : 'next';
				fc.find(`.fc-${ direction }-button`).click();
			},
```


Unfortunately, the "today" button is broken using this hack. But good enough for us to ship.

arshaw:
[Code complete](https://github.com/fullcalendar/fullcalendar/commits/v2.8.0-dev) and slated for v2.8.0 ([wip docs](https://github.com/fullcalendar/fullcalendar/commits/docs-2.0)). Will notify when released.

jquery 3 now works well

arshaw:
Released in [v2.8.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v2.8.0).

avindra:
[...]

PR #3219, opened 2016-06-13
Set JQuery to maximum of 2.1.3 in bower.json

ownaginatious:
I noticed that JQuery 3.0 is automatically pulled when you do `bower install fullcalendar`. There are still a lot of bugs, seeing as JQuery 3.0 is so new, so it seems like it would be a good idea for the time being to lock it to a known supported version.

I chose 2.1.3, as that's what appears to be used in the help doc examples, such as [this](http://fullcalendar.io/views/agendaWeek/) here.

Only found this project recently, so I'm not sure if the latest JQuery 2.x is okay.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -23,7 +23,7 @@
   },
   "copyright": "2016 Adam Shaw",
   "dependencies": {
-    "jquery": ">=1.7.1",
+    "jquery": ">=1.7.1 && <=2.1.3",
     "moment": ">=2.5.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


arshaw:
FullCalendar jquery 3 support is in the works: #3197 and #3124

i want to make FC compatible with jq3 for the next release to avoid needing to force a "legacy" (<=2) version of jq on people. After jq3 compatibility is implemented, a dev can choose do force to jq2 by doing so in their project's bower deps. plz stay tuned

Issue #2433, opened 2015-08-19
Method to retrieve all event sources


arshaw:
Originally reported on Google Code with ID 2168
```
This has been implemented here, via @gmoore:
https://github.com/arshaw/fullcalendar/pull/105

Probably relates to Issue 678
```

Reported by `adamrshaw` on 2014-06-05 22:17:17

Imported with **2** stars.

caseyjhol:
Do we want to implement this similar to #105, simply returning all event sources? Or, do we want to implement this similar to clientEvents, where you can pass through a filter or ID as well?

I realize event sources don't currently officially support having IDs, but if we're going to support refetching specific event sources, allowing them to have IDs might make things more convenient convenient (and wouldn't require any extra code as far as FullCalendar is concerned). Then, we can make the getEventSources API exactly like clientEvents (which would get my vote).

arshaw:
i'd rather have it be more like #105 and only return an array. (i'd actually have it return a *copy* of the internal array to avoid mutations by the caller. learned this the hard way in the past).

I'm hesitant to add an ID system into event sources, and hesitant to add grepping functionality, as I'd prefer that to be the responsibility of the caller. My only argument: to avoid complexity. hopefully the caller can identify their own sources from the array result. no stopping them from putting their own ID in there.

FWIW, i regret having a `clientEvents` method that returns both a single-event and an array of events, depending on the input argument. I plan on breaking out into two methods: `getClientEvent(id)` and `getClientEvents()`

arshaw:
@caseyjhol if you want to make a `getEventSource(id)` method that accepts a single "event source ID", i approve of that. event source IDs on the given source definition objects should be optional of course. plz make separate PR if u do so, thx

caseyjhol:
Ok - I should be able to get those submitted this week.

caseyjhol:
The first item in the `sources` array is always `{ events: [] }`. Since I'm planning on using slice anyway when returning the sources (to return a copy), I'm wondering if we should just slice out that first item? Or is there some potential usage for it that I'm missing?

arshaw:
that initial empty event source is meant for sticky events from [renderEvent](http://fullcalendar.io/docs/event_rendering/renderEvent/).

i would NOT include that event source in the result for `getEventSources`.

ya, slicing out the first item sounds great. can u make sure to write a comment about what's going on?

caseyjhol:
[...]

PR #3223, opened 2016-06-14
add getEventSource method

caseyjhol:
This method retrieves a single event source as discussed in #2433. If there are multiple event sources with same ID, the first event source is retrieved (even though that's technically invalid, as IDs should be unique).

See an example here: http://embed.plnkr.co/HFwUWbNIS5SbeQ1JnZUw/. Clicking the `getEventSource` button adds the event source with ID `source1` to the console.

### DIFF
--- a/tests/automated/getEventSource.js
+++ b/tests/automated/getEventSource.js
@@ -1,0 +1,47 @@
+describe('getEventSource', function() {
+	var options;
+	var calendarEl;
+
+	beforeEach(function() {
+		affix('#cal');
+		calendarEl = $('#cal');
+		options = {
+			now: '2015-08-07',
+			defaultView: 'agendaWeek',
+			eventSources: [
+				{
+					events: [
+						{ id: 1, start: '2015-08-07T02:00:00', end: '2015-08-07T03:00:00', title: 'event A' }
+					],
+					id: 'source1'
+				},
+				{
+					events: [
+						{ id: 2, start: '2015-08-07T03:00:00', end: '2015-08-07T04:00:00', title: 'event B' }
+					],
+					id: 'source2'
+				},
+				{
+					events: [
+						{ id: 3, start: '2015-08-07T04:00:00', end: '2015-08-07T05:00:00', title: 'event C' }
+					],
+					id: 'source3'
+				}
+			]
+		};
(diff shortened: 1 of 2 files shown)


arshaw:
looks great. even tho this is a very simple method, could you please write a test for it (very minimal)?

tests/automated/getEventSource.js
```diff
@@ -0,0 +1,47 @@
...
+
+	it('does not mutate when removeEventSource is called', function(done) {
+		var eventSource;
+
+		calendarEl.fullCalendar(options);
+
+		eventSource = calendarEl.fullCalendar('getEventSource', 'source1');
+		expect(typeof eventSource).toBe('object');
+
+		calendarEl.fullCalendar('removeEventSource', eventSource);
```
arshaw: it would be impossible for this to happen in javascript (can't unset a local reference via side effects in another scope) so could you remove it?

tests/automated/getEventSource.js
```diff
@@ -0,0 +1,47 @@
...
+		};
+	});
+
+	it('does not mutate when removeEventSource is called', function(done) {
+		var eventSource;
+
+		calendarEl.fullCalendar(options);
+
+		eventSource = calendarEl.fullCalendar('getEventSource', 'source1');
+		expect(typeof eventSource).toBe('object');
```
arshaw: would you be able to add to add a check that the object's `id` property is `source1` ?

arshaw: also, would you be able to query for the second source as well, and test for it's `id` as well? I'm anticipating a hypothetical future bug where `getEventSource` erroneously only returns the first source. would happen if the grep statement was written incorrectly.

arshaw:
also good to go. thanks

arshaw:
[Code complete](https://github.com/fullcalendar/fullcalendar/commits/v2.8.0-dev) and [docs written](https://github.com/fullcalendar/fullcalendar/commits/docs-2.0) but not yet released. Will update this thread when released.

I renamed this to `getEventSourceById` to be more explicit.

arshaw:

http://fullcalendar.io/docs/event_data/getEventSourceById/

PR #3239, opened 2016-06-22
Add build badge for Travis to README

avindra:
Build is currently unknown, but will have a status after the next commit to master.


arshaw:
awesome. i'll try to get those tests fixed before merging this, in order to avoid embarrassment

arshaw:
build is now passing. i merged this. thanks
(ps- i had to change the badge url away from sheilds.io (what is that?) to the direct travis-ci.org one because it wasnt loading)

avindra:
The shields.io one lets you customize the style of the badge, and probably does other things too. It's used by some pretty famous libraries (e.g., redux) but it may require some additional setup to work ( I can't remember)

PR #3246, opened 2016-06-25
Added Norwegian Nynorsk translation

zerodogg:
Adds a Norwegian Nynorsk translation. There are already translations for jQuery UI datepicker and moment.js.


arshaw:
thank you!

arshaw:
this is released in v2.9.1, thank you!

PR #3253, opened 2016-06-28
Added error message for unknown method.

xxAtrain223:


src/main.js
```diff
@@ -26,6 +26,9 @@ $.fn.fullCalendar = function(options) {
 					element.removeData('fullCalendar');
 				}
 			}
+			else {
+				console.error("'" + options +"' is an unknown FullCalendar method.");
```
arshaw: i would rather remove the isFunction check above, and have a real exception be thrown, instead of doing console errors. better because of stack trace

arshaw:
Released in v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/#more-257))

PR #3266, opened 2016-07-08
Travis: add style checks to CI and speed up runs

avindra:
This should speed up subsequent runs by caching `node_modules` and `lib`.

Additionally, this adds linting to Travis, and fixes existing lint problems in the tests.

----

Example run: https://travis-ci.org/avindra/fullcalendar/builds/143421961

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: node_js
 node_js:
   - 'node' # implies latest version
 
+cache:
+  directories:
+  - node_modules
+  - lib
+
 env:
   - CMD=test
   - CMD=lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,11 @@ language: node_js
 node_js:
   - 'node' # implies latest version
 
+env:
+  - CMD=test
+  - CMD=lint
+
 before_script:
   - bower install
+
+script: npm run $CMD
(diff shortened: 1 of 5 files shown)


arshaw:
awesome, i'll merge this

arshaw:
@avindra, unrelated question: would you be able to help get this exact TravisCI setup going for [fullcalendar-scheduler](https://github.com/fullcalendar/fullcalendar-scheduler)? would be really cool

.travis.yml
```diff
@@ -2,5 +2,16 @@ language: node_js
...
+  - lib
+
+env:
+  - CMD=test
+  - CMD=lint
+
 before_script:
   - bower install
+
+script: npm run $CMD
```
arshaw: what does this line do?

avindra: Script defines any custom script you want travis to run. For each env line, it will trigger a separate build basically, with the environment variable you specify in the line. So basically this configuration does `npm run test` and `npm run lint` instead of just `npm test`. They will run in parallel and each have their own build output, which is better than just doing something like `lint && test`.

avindra:
@arshaw : PR made @ https://github.com/fullcalendar/fullcalendar-scheduler/pull/195

PR #3271, opened 2016-07-11
allow height/contentHeight to accept a function for dynamic value

caseyjhol:
This is a fork of #3040. Instead of adding a new `getContainerHeight` option, this simply allows height and contentHeight to accept a function.

See a preview here: http://plnkr.co/edit/O0iBHMcFa0faPE8XyPb3?p=preview where `height` is a function that returns the height of the calendar's parent (which is an absolutely positioned DIV with `{top: 200px, right: 0, bottom: 0, left: 0;}`).

### DIFF
--- a/tests/automated/height-and-contentHeight.js
+++ b/tests/automated/height-and-contentHeight.js
@@ -1,11 +1,19 @@
 (function() {
 
-[ 'height', 'contentHeight' ].forEach(function(heightProp) { describe(heightProp, function() {
+[ 'height', 'contentHeight' ].forEach(function(heightProp) { ddescribe(heightProp, function() {
 
 	var calendarEl;
 	var options;
 	var heightElm;
 	var asAMethod;
+	var heightPropDescriptions = [
+		{ description: 'as a number', height: 600 },
+		{ description: 'as a function', height: getParentHeight, heightWrapper: true }
+	];
+
+	if (heightProp === 'height') {
+		heightPropDescriptions.push({ description: 'as "parent"', height: 'parent', heightWrapper: true });
+	}
 
 	beforeEach(function() {
 		affix('#cal');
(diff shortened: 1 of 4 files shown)


tests/automated/height-and-contentHeight.js
```diff
@@ -108,6 +114,61 @@
 				});
 			});
 
+			describe('as a function, when there are no events', function() {
```
arshaw: not a big deal, but this is a lot of copypasta from the other tests. would you be able to DRY it up? recommended format to do so:

```js
[
	{ description: 'as a number', height: 600 },
	{ description: 'as a function', height: getParentHeight, heightWrapper: true }
].forEach(function(testInfo) {

	describe(testInfo.description, function() {

		if (testInfo.heightWrapper) {
			beforeEach(function() {
				calendarEl.wrap('<div class="calendar-container" style="height: 600px;" />');
			});
		}

		describe('when there are no events', function() {
			// ...
		}); 
		// ...
	});
});
```

arshaw: also, this will make a `"parent"` setting much easier to write tests for (pretty much the same tests as `getParentHeight`)

arshaw:
awesome! thanks for carrying this through to the finish line. would you be able to add another setting as a shortcut to setting a 100% to the parent? the value would just be `'parent'`.

One thing we *might* want to test for is that the function and `'parent'` args get updated upon window resize. Making an automated test for this will be too cumbersome and inaccurate. But would you mind making a *manual* test for this?

Could you just modify this existing manual test?
https://github.com/fullcalendar/fullcalendar/blob/master/tests/fullheight.html
Will be a great proof of concept!

arshaw:
looks great! thanks for the hard work on this one. i'll merge it into the upcoming release.

caseyjhol:
In my testing, the height does get auto-updated when the window is resized, but not as smoothly as when doing `$(window).resize` (i.e. it updates less frequently). It looks like that's because `windowResizeDelay` is set to 200ms. I can manually override that to 0, but it's not documented.

arshaw:
I'll document `windowResizeDelay` also make the default value significantly smaller

arshaw:
this is released in v2.9.1

docs i've updated:
http://fullcalendar.io/docs/display/height/
http://fullcalendar.io/docs/display/contentHeight/

windowResizeDelay:
http://fullcalendar.io/docs/display/windowResizeDelay/

thank you @caseyjhol !

aharpervc:
cool 👍

Issue #3232, opened 2016-06-20
Event limit popover clipped by overflow


arshaw:
![screen shot 2016-06-19 at 9 47 14 pm](https://cloud.githubusercontent.com/assets/56520/16183507/7f6e821e-3667-11e6-9cd8-ccb0d363a179.png)

KrunchMuffin:
Worse if at bottom of calendar.
![image](https://cloud.githubusercontent.com/assets/1037197/16198917/9eaad8a8-36d5-11e6-8699-e70f91426fae.png)

SirDre:
@arshaw I think, I solve this issue. Please confirm if you like this change.

The following snip below are the changes you would make within the views: 
![pullover_fix](https://cloud.githubusercontent.com/assets/1511532/16422401/7a70940a-3d15-11e6-83f7-821cc6aae49f.png)

Also, within the css:
![pullover_fix_css](https://cloud.githubusercontent.com/assets/1511532/16422658/6f7c2266-3d16-11e6-8981-f09449cc2f51.png)

Preview example:
http://cifullcalendar.com/v2/

fluciano:
Hi Adam,
Do you have a time frame on when this issue (Event limit popover clipped by overflow) will be resolved?

arshaw:
this has been fixed and release in [v2.9.1](https://github.com/fullcalendar/fullcalendar/releases/tag/v2.9.1).

KiranContentserv:
I have updated with the latest version of full-calendar but still its clipping. For the browser viewport its working but i have loaded fullcalendar in indvidual div.
![capture](https://user-images.githubusercontent.com/45969301/50154903-a4b14b80-02f0-11e9-99ea-603f25374c30.PNG)

PR #3286, opened 2016-07-28
popover option for centered and limits 

benitogf:
this could be a solution for https://github.com/fullcalendar/fullcalendar/issues/3232 ?


arshaw:
@benitogf, no not really. one is a feature the other is a bug

arshaw:
I'm not sure how many others want this, and if making this a core setting is justified. Could you make a normal issue ticket (not a PR)? Doing this would allow people to vote up the issue. It would be great if you could put a screenshot in the ticket as well.

PR #3299, opened 2016-08-03
Add a footer option

axelduch:
@arshaw I ensured I did not break unit tests. Even though currently there seem to be 6 tests failing (in the original master branch).


rahulthewall:
Hi, 

I need to duplicate the header as my footer in the project that I am doing. Other than waiting for the PR to be merged, is there any other way I can `hack` a workaround for now?

goatlady102:
So we are discussing the same portion -  send me a screen shot with the area you want duplicated circled

rahulthewall:
Here's the screenshot: http://imgur.com/a/11QHb

src/Calendar.js
```diff
@@ -738,25 +781,26 @@ function Calendar_constructor(element, overrides) {
...
 			suggestedViewHeight = Math.round(content.width() / Math.max(t.options.aspectRatio, .5));
 		}
 	}
 
 
-	function queryHeaderHeight() {
-		return header.el ? header.el.outerHeight(true) : 0; // includes margin
+	function queryToolbarsHeight() {
+		var firstToolbar = toolbarsManager.items[0];
+		return firstToolbar.el ? firstToolbar.el.outerHeight(true) : 0; // includes margin
```
arshaw: we'd actually want to get the *total cumulative* height of the toolbars. you can test with a demo like this: https://github.com/fullcalendar/fullcalendar/blob/master/tests/fullheight.html (has 100% height)

arshaw:
looks good!

could you write some very basic tests for this?

1. that the `height` setting is not thrown off by this new toolbar (related to my inline comment)
2. that it simply renders. maybe just one test. a very minimal version of this: https://github.com/fullcalendar/fullcalendar/blob/master/tests/automated/header-rendering.js
3. that it accepts clicks and responds. maybe just one test. a very minimal version of this: https://github.com/fullcalendar/fullcalendar/blob/master/tests/automated/header-navigation.js

axelduch:
I had a look at travis, the checks that fail are those I was talking about at the very beginning of this PR.

axelduch:
Also, for the `height` check, what should be checked precisely? I'm not so comfortable with this part, to be honest.

arshaw:
the main thing that should be checked: when you set the `height` option to `"parent"` (see http://fullcalendar.io/docs/display/height/) the calendar should fill the entire height of the container. there's a manual test for that:
https://github.com/fullcalendar/fullcalendar/blob/master/tests/fullheight.html

it works by setting the height if the view's div, while still making the total height of the outer calendar's container match the desired parent height. it must take the total parent height and subtract anything that is not the view, such as toolbars.

does that help?

arshaw:
some intro to automated tests:

some automated tests:

axelduch:
@arshaw I was uncomfortable with this particular part of tests, but thanks for sharing! And yes it did help a lot, I know what is expected, I'll write some tests for it.

But still this is kind of weird, the 6 tests failing, actually I was surprised at the very beginning of this PR that Travis build passed, because it did not pass locally but since I ran the test on fullcalendar/fullcalendar master branch too to make sure that it was not one of my changes which broke the unit tests, I took a chance.

And then, once I pushed two days ago, Travis actually failed with the same output I had on fullcalendar/fullcalendar master branch originally. Do you have an idea on what's happening here? Because I don't think this is in the scope of this issue. Even though I'd be happy to help about it.

arshaw:
what tests are failing?

could you try merging in [v3.0-beta](https://github.com/fullcalendar/fullcalendar/tree/v3-beta) and see if that resolves? that will be the next release that goes out

caseyjhol:
@arshaw You should be able to see the failed tests here: https://travis-ci.org/fullcalendar/fullcalendar/jobs/152216920

tests/fullheight.html
```diff
@@ -9,18 +9,23 @@
 <script>
 
 	$(document).ready(function() {
-	
+
 		var date = new Date();
 		var d = date.getDate();
 		var m = date.getMonth();
 		var y = date.getFullYear();
-
```
arshaw: any way you can stop your text editor from deleting these lines?

axelduch: Oh trailing the whitespaces, yes I won't push'em anymore, sorry about that. I'll open a PR on v3.0-beta and see if the travis build passes.

axelduch:
I have merged locally my branch to v3-beta, anything I should know about the build now that gulp replaced grunt? So far Toolbar.js does not seem to be added to the minify step, so I get errors with Toolbar being `undefined`.

axelduch:
@arshaw Seems like it no longer complains during the build/testing phase, do you think it is okay? I tested manually the height with the provided manual test that I updated.

arshaw:
would still really like automated tests. otherwise, I (or someone else) would have to write them. they are mostly important to prevent regressions in the future.

axelduch:
@arshaw I already did (cf. https://github.com/fullcalendar/fullcalendar/pull/3299/commits/70ad5b3f402c205ec11aec42aae5e73d7ee4471d https://github.com/fullcalendar/fullcalendar/pull/3299/commits/b86601f1c951607db86dd9f7ffa9b712427a45b0)

arshaw:
oh ok, sorry, i misread your comments then. i'll check this out this weekend!

src/common/Iterator.js
```diff
@@ -0,0 +1,25 @@
+function Iterator(items) {
+    this.items = items || [];
+}
+
+
+Iterator.prototype.iterate = function(callback) {
```
arshaw: i saw you use `map` before on the `items` internal array... in the same vein, could we use `forEach` directly on the `items` array?

src/Calendar.js
```diff
@@ -590,15 +598,50 @@ function Calendar_constructor(element, overrides) {
...
+	function computeHeaderOptions() {
+		return $.extend(
+			{ extraClasses: 'fc-header' },
+			t.options,
+			{ layout: t.options.header });
+	}
+
+
+	function computeFooterOptions() {
+		return $.extend(
```
arshaw: i'd prefer not to mix the external api options (`t.options`) with toolbar-specific options because there might be collisions, and they are fundamentally different.

For accessing calendar-object-based API options, could you access `calendar.options` directly (in Toolbar.js) and for toolbar-specific options, rename it to something like `toolbarOptions` and `setToolbarOptions` to indicate they are different? (in Toolbar.js)

src/Calendar.js
```diff
@@ -590,15 +598,50 @@ function Calendar_constructor(element, overrides) {
...
+		return [
+			new Toolbar(t, computeHeaderOptions()),
+			new Toolbar(t, computeFooterOptions())
+		];
+	}
+
+
+	function computeHeaderOptions() {
+		return $.extend(
+			{ extraClasses: 'fc-header' },
```
axelduch: Sure, makes sense

src/common/Iterator.js
axelduch: Yeah seems like I got carried away here, nice one.

axelduch:
@arshaw Actually I was thinking, shouldn't we leave the `.fc-header` class for backward compatibility? There are probably a bunch of people out there redefining styles for it, what do you think?

arshaw:
@axelduch where do you see the `fc-header` className being used in the current live code?

axelduch:
@arshaw There are [a few places](https://github.com/fullcalendar/fullcalendar/search?utf8=%E2%9C%93&q=fc-header) but apart from testing files, `common.css` and `DayGrid.limit.js`.

arshaw:
yes, the `fc-header` is used by other parts of the rendered view, but since it was never used to identify the toolbar area, i don't see any backward compatibility concerns. From my POV, we are taking the element that was previously `fc-toolbar` and giving it another className: `fc-header-toolbar`. Yes, the "footer" will now have the same className (`fc-toolbar fc-footer-toolbar` to be precise), but I wouldn't consider this a breaking change because the footer toolbar is a new feature that nobody has previously used.

axelduch:
Then should be good to go if you agree with the latest changes I made. :)

src/common/common.css
```diff
@@ -40,12 +40,12 @@ body .fc { /* extra precedence to overcome jqui */
 }
 
 .fc-unthemed .fc-divider,
-.fc-unthemed .fc-popover .fc-header,
+.fc-unthemed .fc-popover .fc-header-toolbar,
```
arshaw: the `fc-header` that lives in the `fc-popover`, i'd like to keep that classname as-is. i don't see a reason to change it.

arshaw: it has nothing to do with toolbars

src/common/Iterator.js
arshaw: can you do away with this method entirely and just call `this.items.forEach` instead? only called in once place

src/common/common.css
axelduch: Oh my bad sorry I'll restore this

arshaw:
https://fullcalendar.io/docs/display/footer/
thanks!

PR #3303, opened 2016-08-08
Add .fc scope to table print styles

juodumas:
Fullcalendar's print stylesheet was overriding table styles globally. This commit scopes the rules.

### DIFF
--- a/src/common/print.css
+++ b/src/common/print.css
@@ -32,11 +32,11 @@
 /* Table & Day-Row Restyling
 --------------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


arshaw:
thanks! looks good. i'm just wondering if the precedence change will unintentionally begin to override other statements in the print stylesheet. how exhaustively have you tested? tested in both month/agenda view?

juodumas:
I have just compared month/agendaDay view printouts before and after the change and they match.
Created PR#3343 for 3.0.0, closing this one.

PR #3310, opened 2016-08-16
Add lang option rename to changelog (v3-beta)

peternowee:
The only thing that surprised me in v3-beta was the renaming of the `lang` option to `locale`. Assuming you do not want to support `lang` as a deprecated alternative name, perhaps this change can be listed with the Breaking Changes in the Changelog (now it only mentions the renaming of the dist/lang-files).


arshaw:
thanks for this. i forgot to include this explicitly

Issue #1009, opened 2015-08-19
Icons on event elements


arshaw:
Originally reported on Google Code with ID 739

```
It would be nice to be able to append an icon preceding/following the time in the event
time span. This would be used, for example, to have an icon indicating an event repeats,
or has an alert. 

My thought is to provide an icon and location option that can be appended to each event
based on one or more types defined in the Event properties. 

Or possibly the ability to add an additional class to the time span based on an Event
type code. The additional class could contain a background image and position. 

Much appreciated!
```

Reported by `pwc1011` on 2010-11-30 19:20:56

Imported with **5** stars.

MartijnWelker:
Not sure if this is still active but +1

StudioJunkyard:
This might give you an idea of what I do to achieve something along these ideas.
This is using jQuery v2.1.1, Bootstrap v3.3.6 + Bootstraps JS for the popovers. The icons are pulled in using a PHP function to retreive the SVG Icons (for e.g. `<?php svg('edit');?>`).
I've only included the part that does mouseover to show the icons, and popover with the extra information.

```css
#events-layer{position:absolute;top:-1px;right:0;z-index:100;width:65px;height:17px;text-align:right;padding-right:5px}
#events-layer .btn{padding:0;line-height:13px;font-size:12px}
#events-layer .btn-xs{width:14px;height:14px}
```

```javascript
eventMouseover:function(event,domEvent,view){
                    var layer='<div id="events-layer" class="fc-transparent">';
                    if(event.status=="unconfirmed")layer+='<span id="cbut'+event.id+'" class="btn btn-default btn-xs"><?php svg('approve');?></span> ';
                    layer+='<span id="edbut'+event.id+'" class="btn btn-default btn-xs"><?php svg('edit');?></span> <span id="delbut'+event.id+'" class="btn btn-default trash btn-xs"><?php svg('trash');?></span></div>';
                    var content='Start: '+$.fullCalendar.moment(event.start).format('HH:mm');
                    if(event.end>event.start)content+='<br>End: '+$.fullCalendar.moment(event.end).format('HH:mm');
                    if(event.description!='')content+='<br>'+event.description;
                    $(this).append(layer);
                    var el=$(this);
                    $("#cbut"+event.id).click(function(){
                        $("#cbut"+event.id).remove();
                        $("#events-layer").remove();
                        event.color="#5cb85c";
                        event.status="confirmed";
                        update(event.id,"content","status","confirmed");
[...]

PR #3342, opened 2016-09-10
Allow the event to add classes to the title span.

vanboom:
This enables the ability to prefix an icon to the event using CSS ::before. 

The event can now define a "titleClass" to be added to the event title span. 

See #1009

### DIFF
--- a/src/common/DayGrid.events.js
+++ b/src/common/DayGrid.events.js
@@ -114,7 +114,7 @@ DayGrid.mixin({
 		}
 
 		titleHtml =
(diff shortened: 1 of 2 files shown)


arshaw:
why couldn't you just set the `className` of the [Event Object](https://fullcalendar.io/docs/event_data/Event_Object/) and then target the title using a nested selector. for example:

```
JS:
events: [
    { title: 'my event', className: 'customeventclass', blablabla }
]

CSS:
.customeventclass .fc-title::before {
}
```

PR #3343, opened 2016-09-13
Add .fc scope to table print styles

juodumas:
Fullcalendar's print stylesheet was overriding table styles globally. This commit scopes the rules.

Updated #3303 for 3.0.0.
I have compared month/agendaDay view printouts before and after the change and they match.


arshaw:
thanks!

arshaw:
release in [v3.0.1](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.0.1)

Issue #2938, opened 2015-11-12
Using debounce in renderEvents


melwas:
Hi, 

Have you thought about using debounce function when rendering events? I did a really quick test and replaced `renderEvents` with debounced version (in `reportEvents`) and performance improvement is significant (e.g. when adding and removing eventSources).

``` javascript
var renderEventsDebounced = debounce(renderEvents, 50);
```

I am just curious, because maybe you have tried this before but discovered some important disadvantages?

arshaw:
this is a route i want to go down. this is a solution to #2681 an other issues where people want to avoid needless event rerendering. the downside is that event rendering becomes asynchronous, so it might make it difficult for people to know when it's finished, but that's where callbacks like [eventAfterAllRender](http://fullcalendar.io/docs/event_rendering/eventAfterAllRender/) come in.

laoneo:
+1

arshaw:
would help with "performance issues" #2524

M0nter0:
Thanks a lot, @melwas 
It's a great solution. In my case, we were adding 100 slots at once, freezing the browser. Now using debounce() with 50 miliseconds they appear "instantly" without problems.

kferrio:
+1

rwwagner90:
Would love to be able to use this solution, could someone point me to where I need to put this debounce in? Is it here? https://github.com/fullcalendar/fullcalendar/blob/master/src/Calendar.js#L274

arshaw:
you'd likely want to debounce the following lines: https://github.com/fullcalendar/fullcalendar/blob/v2.9.1/src/Calendar.js#L792-L794

lunayo:
how should I change the code? I tried this but the performance is still quite slow
```
function renderEvents() { // destroys old events if previously rendered
  if (elementVisible()) {
	//freezeContentHeight();
	//currentView.displayEvents(events);
	//unfreezeContentHeight();
	debounce(currentView.displayEvents(events), 50);
   }
}
```

jamesdixon:
@arshaw is this on the roadmap for a 2.x release or are you looking at overall performance improvements being part of 3.x?

arshaw:
@lunayo you're using debounce incorrectly. it should be something like this:
```js
var debouncedDisplayEvents = debounce(function() {
	freezeContentHeight();
	currentView.displayEvents(events);
	unfreezeContentHeight();
}, 50);

function renderEvents() { // destroys old events if previously rendered
[...]

Issue #2524, opened 2015-08-19
Performance decrease in 2.x


arshaw:
Originally reported on Google Code with ID 2259

```
The following jsFiddles use the same code, only switching from 2.0.0 to 2.1.0 versions
of FullCalendar.  On a Core i7 FireFox the 2.0.0 versions loads 1500 events in under
a second, the 2.1.0 version takes 3 - 4 seconds.  Also the removeEventSource is extremely
slow.


2.0.0 - http://jsfiddle.net/indexic/u6youn1b/
2.1.0 - http://jsfiddle.net/indexic/L9z2Lbur/
```

Reported by `google@indexic.net` on 2014-08-27 20:55:20

Imported with **14** stars.

xxAtrain223:
I'm sure you already know this, but i've done some digging and at least one of the issues is with displayEvents, specifically how it [clears and renders events](https://github.com/fullcalendar/fullcalendar/blob/master/src/common/View.js#L614-L615). This is the brute force approach, and breaks down when there are about 50 events, causing about half a second delay per update event. With one update, it's not that bad, but with several updates...

Again, i'm sure you already know all of this.

Funny thing, this made me write more efficient, optimized code.

arshaw:
hi @xxAtrain223, using `debounce` is the leading proposed solution to this:
https://github.com/fullcalendar/fullcalendar/issues/2938 (not yet implemented)
&uarr;
Could you please visit the link, **give a** 👍  **to vote** ([?](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments)), press the 🔈**Subscribe** button to receive updates?

essentially throttling rerenders. you'll likely want to implement this logic on your app's end in the meantime. (sounds like you might have already)

rwwagner90:
@arshaw could you tell us how to implement the debouncing? I see in the issue that it's supposed to go in reportEvents I think, but that doesn't seem to be a function, do I define that on my side?

Manuweb2:
Hi @arshaw,

Thanks for the amazing job on fullcalendar.
We also noticed that using more than 600 events on a week calendar (yes we have clients with buzy weeks...) it took the calendar about 4s second to display the elements.
We are using a single source with an ajax call for the source using event: url and the ajax return a json.
The ajax call is returned in about 200ms, then we can count to 4s until the elements appear on the calendar... we were using the 2.0.0beta until now and we thought updating to 3 would change that but it has no effect and still 4s until the elements appear. 
[...]

PR #3351, opened 2016-09-19
Add batch-render methods: renderEvents and updateEvents

MartijnWelker:
`batchRenderEvent`/`batchUpdateEvent` will be a substitute for `renderEvent`/`updateEvent` when you have more then one event to render/update.

Input: `arrayOfEvents, stick`
Output: `arrayOfRenderedEvents`

`renderEvent` will call `reportEvents` every call, so calling it in a loop is very expensive. I have created `batchRenderEvent` to overcome this.

Speed comparison: http://jsbin.com/gugowemivu/edit?js,console,output

I didn't make an automated test because `renderEvent` also doesn't have one.

Possible fix for: #2938 
Helps for: #2524

### DIFF
--- a/src/EventManager.js
+++ b/src/EventManager.js
@@ -12,8 +12,8 @@ var eventGUID = 1;
 
 function EventManager() { // assumed to be a calendar
 	var t = this;
-	
-	
+
+
 	// exports
 	t.isFetchNeeded = isFetchNeeded;
 	t.fetchEvents = fetchEvents;
@@ -27,17 +27,18 @@ function EventManager() { // assumed to be a calendar
 	t.removeEventSources = removeEventSources;
 	t.updateEvent = updateEvent;
 	t.renderEvent = renderEvent;
+	t.batchRenderEvent = batchRenderEvent;
 	t.removeEvents = removeEvents;
 	t.clientEvents = clientEvents;
 	t.mutateEvent = mutateEvent;
 	t.normalizeEventDates = normalizeEventDates;
 	t.normalizeEventTimes = normalizeEventTimes;
-	
-	
+
+
 	// imports
 	var reportEvents = t.reportEvents;
-	
-	
+
+
 	// locals
 	var stickySource = { events: [] };
 	var sources = [ stickySource ];
@@ -55,9 +56,9 @@ function EventManager() { // assumed to be a calendar
 			}
 		}
 	);
-	
-	
-	
+
+
+
 	/* Fetching
 	-----------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


arshaw:
@MartijnWelker, what's the problem with calling `reportEvents` frequently? re-rendering of all events? I'm trying to understand how this solutions related to using `debounce` for rendering events: https://github.com/fullcalendar/fullcalendar/issues/2938

MartijnWelker:
@arshaw This is more as an alternative for the `debounce` option to prevent making rendering events async. The problem with `reportEvents` is that it will render all the events, which is painfully slow in a loop (as can be seen in the jsbin)

src/EventManager.js
```diff
@@ -26,18 +26,20 @@ function EventManager() { // assumed to be a calendar
 	t.removeEventSource = removeEventSource;
 	t.removeEventSources = removeEventSources;
 	t.updateEvent = updateEvent;
+	t.batchUpdateEvent = batchUpdateEvent;
```
arshaw: could we name the method to `updateEvents` (plural)?

src/EventManager.js
```diff
@@ -26,18 +26,20 @@ function EventManager() { // assumed to be a calendar
...
 	t.renderEvent = renderEvent;
+	t.batchRenderEvent = batchRenderEvent;
```
arshaw: could we name the method to `renderEvents` (plural)?

src/EventManager.js
```diff
@@ -503,6 +505,30 @@ function EventManager() { // assumed to be a calendar
 	}
 
 
+	// Only ever called from the externally-facing API
+	function batchUpdateEvent(events) {
```
arshaw: instead of having this function share lots of repeat logic with `updateEvent`, could you unify their code? have just one function that handles the array case, and then the `updateEvent` method would leverage that with a single-item array.

src/EventManager.js
```diff
@@ -552,8 +578,43 @@ function EventManager() { // assumed to be a calendar
 
 		return [];
 	}
-	
-	
+
+
+	// returns the expanded events that were created
+	function batchRenderEvent (eventInput, stick) {
```
arshaw: same note about unifying this with the single-item version of this function

arshaw:
looks great, thanks! I'll try my best to get this into the next release.

arshaw:
https://fullcalendar.io/docs/event_rendering/renderEvents/
https://fullcalendar.io/docs/event_data/updateEvents/
thanks!

PR #3365, opened 2016-09-29
implement fc-today in agendaWeek day header

### LINKED ISSUE #3361: fc-today missing in agendaWeek day header
### Bug Reports

http://jsbin.com/hahuzes/6/edit?html,js,output
### Feature Requests

Is it possible to add the fc-today class to the header like the day header have in the month view?

Im not sure if its wanted to have the background of the header to be like cells but the css could be adjusted.
If not acceptable, i guess i can do exactly like in the jsbin and add it manually.

Was mixed in #3337

Rouche:
https://github.com/fullcalendar/fullcalendar/issues/3361

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ dist
 # npm
 node_modules
 npm-debug.log
+/.settings/
+/.project
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,3 @@ dist
 # npm
 node_modules
 npm-debug.log
-/.settings/
-/.project
--- a/src/common/DayTableMixin.js
+++ b/src/common/DayTableMixin.js
@@ -56,7 +56,7 @@ var DayTableMixin = FC.DayTableMixin = {
 		this.dayIndices = dayIndices;
 		this.daysPerRow = daysPerRow;
 		this.rowCnt = rowCnt;
-		
+
 		this.updateDayTableCols();
 	},
 
@@ -295,8 +295,10 @@ var DayTableMixin = FC.DayTableMixin = {
 	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
 		var view = this.view;
 
+        var todayClass = date.isSame(view.calendar.getNow(), 'day') ? ' fc-today' : '';
+
 		return '' +
-			'<th class="fc-day-header ' + view.widgetHeaderClass + ' fc-' + dayIDs[date.day()] + '"' +
+            '<th class="fc-day-header ' + view.widgetHeaderClass + todayClass + ' fc-' + dayIDs[date.day()] + '"' +
 				(this.rowCnt === 1 ?
 					' data-date="' + date.format('YYYY-MM-DD') + '"' :
 					'') +
--- a/src/common/DayTableMixin.js
+++ b/src/common/DayTableMixin.js
@@ -295,10 +295,11 @@ var DayTableMixin = FC.DayTableMixin = {
 	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
 		var view = this.view;
 
-        var todayClass = date.isSame(view.calendar.getNow(), 'day') ? ' fc-today' : '';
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@ dist
 # npm
 node_modules
 npm-debug.log
+/.settings/
```
espen: these should probably be removed

Rouche: Its for eclipse config. It regenerate automatically on import.

espen: But how is that relevant for this project?

Rouche: Ill remove it. Its not for the project, thats why its on ignore.
Just usefull for devs using eclipse. Like ignoring node_modules / dist / etc.

espen: dist and node_modules are generated by Fullcalendar so that's why they are included here. For anything else I recommend using a global gitignore for env-specific ignores.

src/common/DayTableMixin.js
```diff
@@ -295,8 +295,10 @@ var DayTableMixin = FC.DayTableMixin = {
 	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
 		var view = this.view;
 
+        var todayClass = date.isSame(view.calendar.getNow(), 'day') ? ' fc-today' : '';
+
 		return '' +
-			'<th class="fc-day-header ' + view.widgetHeaderClass + ' fc-' + dayIDs[date.day()] + '"' +
+            '<th class="fc-day-header ' + view.widgetHeaderClass + todayClass + ' fc-' + dayIDs[date.day()] + '"' +
```
arshaw: can you correct this whitespace change? should be tabs

src/common/DayTableMixin.js
```diff
@@ -295,8 +295,10 @@ var DayTableMixin = FC.DayTableMixin = {
 	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
 		var view = this.view;
 
+        var todayClass = date.isSame(view.calendar.getNow(), 'day') ? ' fc-today' : '';
```
arshaw: can we start to leverage `getDayClasses` instead?
https://github.com/Rouche/fullcalendar/blob/v3.0.1/src/common/Grid.js#L620

arshaw: also, we don't want the today class to always be there... it won't make sense for a month view, or any other view that has more than one row of days.

can you adjust so that the today class only gets applied when [rowCnt](https://github.com/Rouche/fullcalendar/blob/v3.0.1/src/common/DayTableMixin.js#L12) is `1` ?

Rouche: Mmm Im pretty sure i verified that. I tough this code was not called in month view. Ill double check.

Rouche: Ok your right for the rowCount.

For the ```getDayClasses``` i am really unsure of side effects. It will add ```fc-past fc-future fc-other-month fc-[dayIDs]``` also. And if we use rowCount, we need ot put ```'fc-' + dayIDs[date.day()]``` in a else or the class will be doubled.

is ```view.dayGrid``` the correct way to access the method?

Waiting advise on this one.

Rouche: Perfect with me. Give me a day or 2 and i adjust the commit.

Rouche: Done

arshaw:
looks good.

arshaw:
TODO: look at how these changes affect styling of headers

Rouche:
Style: On the agendaWeek header, when fc-today is applyed, the background gets the same light yellow as the column.
This is the only one i know of.

Rouche:
Mmm, this is weird.
unable to minify?
this was working fine like when i committed earlier.

Lint is working ok too on my side.

arshaw:
released in [v3.1.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.1.0) with an alternate implementation. thanks for the code submission nonetheless!

Issue #3259, opened 2016-07-04
Different longPressDelay according to the type of interaction


Marco-Guido:
It would be very nice to have a way to differentiate the longPressDelay according to the type of interaction on touch devices.
We would like to update to 2.7+ and enable drag-n-drop and resize but can't because our users are used to short delays when selecting a time-range.

arshaw:
thanks for the request. this shouldn't be very hard to implement

jorisvh:
Would you have any time frame in mind to implement this? I'm working with Marco and, as he said, we can't update to the latest versions (we tried and had to rollback because our users are so used to different delays for different actions). If time is the constraining resource, we might be able to help, with a bit of guidance from your side.

caseyjhol:
Could possibly even just add support for an array or object with the same longPressDelay attribute for backwards compatibility. If just a number is used, the same delay is used for all interactions.

arshaw:
hey @caseyjhol i'd rather create new properties altogether, `eventLongPressDelay` and `selectLongPressDelay`, similar to how the `*Constraint` properties are named.

jorisvh:
Hey guys, any news regarding this? @caseyjhol if you already have some code (v2 or v3 are both fine), I'd have no problem patching my files.

caseyjhol:
@arshaw Do you want to keep the longPressDelay property, and have eventLongPressDelay and selectLongPressDelay default to that value? Or do you want to completely remove the longPressDelay property?

arshaw:
@caseyjhol, the former: `eventLongPressDelay` and `selectLongPressDelay` should default to `longPressDelay` if not defined.

toxaq:
I just upgraded to 2.9 from 2.3 after purchasing the scheduler and this long hold was an unexpected change for me. The new properties @arshaw has suggested sound great.

arshaw:
released in [v3.1.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.1.0)
https://fullcalendar.io/docs/event_ui/eventLongPressDelay/
https://fullcalendar.io/docs/event_ui/selectLongPressDelay/

toxaq:
For future visitors, that second link is broken, instead: https://fullcalendar.io/docs/selection/selectLongPressDelay/

PR #3377, opened 2016-10-12
add support for eventLongPressDelay and selectLongPressDelay properties

caseyjhol:
See #3259.

### DIFF
--- a/src/common/Grid.events.js
+++ b/src/common/Grid.events.js
@@ -305,13 +305,14 @@ Grid.mixin({
 		}
 
 		if (!isResizing && (isDraggable || isResizable)) { // allowed to be selected?
+			var eventLongPressDelay = this.view.opt('eventLongPressDelay');
 
 			dragListener = isDraggable ?
 				this.buildSegDragListener(seg) :
 				this.buildSegSelectListener(seg); // seg isn't draggable, but still needs to be selected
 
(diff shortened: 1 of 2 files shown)


src/common/Grid.events.js
```diff
@@ -311,7 +311,7 @@ Grid.mixin({
 				this.buildSegSelectListener(seg); // seg isn't draggable, but still needs to be selected
 
 			dragListener.startInteraction(ev, { // won't start if already started
-				delay: isSelected ? 0 : this.view.opt('longPressDelay') // do delay if not already selected
+				delay: isSelected ? 0 : this.view.opt('eventLongPressDelay') || this.view.opt('longPressDelay') // do delay if not already selected
```
arshaw: can  you make it so that if the `eventLongPressDelay` value is `0`, it WON'T fallback to `longPressDelay` ?

we'll want to do a non-null/undefined test via `!= null`

same for `selectLongPressDelay`.

caseyjhol: Good call. Will get that fixed.

arshaw:
looks good

arshaw:
https://fullcalendar.io/docs/event_ui/eventLongPressDelay/
https://fullcalendar.io/docs/event_ui/selectLongPressDelay/
thanks!

PR #3384, opened 2016-10-21
fixes dynamic height change when multiple options specified

### LINKED ISSUE #3327: contentHeight, height and aspectRatio dynamic change won't work when you set multiple options
Hello,

contentHeight, height and aspectRatio can't be set dynamically with others options. For example :
`$('#calendar').fullCalendar('option', {
    contentHeight: 500,
    anyOtherOption: 'foo'
});`
will **NOT work**.

But
`$('#calendar').fullCalendar('option', {
    anyOtherOption: 'foo'
});
$('#calendar').fullCalendar('option', {
    contentHeight: 500
});`
will **work**.

You can try it here:
http://jsbin.com/vazilaw/edit?js,output

paw3lx:
I fixed the #3327  bug and added some tests (manual and automated).

This is my first ever contribution so I hope it will be ok.

I was using fullcalendar for a long time and now I would like to contribute and help with some things. I can help with fixings bugs, writing some tests or adding features. Can I find some issue and work with it (like I did with this one)?


arshaw:
thanks for the thorough test case and thorough PR.

the way the code works (before you touched it), if there is more than 1 option change, then it always falls back to `reinitView()`.

i would prefer to get `reinitView` rendering the correct height rather than attempting to make `setOptions` smarter. Does that make sense?

tests/manual/issue_3327.html
```diff
@@ -0,0 +1,124 @@
+<!DOCTYPE html>
```
arshaw: i'm sure this file was useful for debugging, but can your remove it from the PR? I believe the automated tests take care of this,

paw3lx: Ok. I will remove it.

paw3lx:
Ok. So let me understand this right. You don't want to touch the setOptions method. I agree with you. 
So now do you want me to make some changes inside reinitView instead of in setOptions (to render the correct height)? 
I can do this, I'm just asking if I understood everything good.

src/Calendar.js
```diff
@@ -669,6 +669,7 @@ function Calendar_constructor(element, overrides) {
 			) {
 				if (elementVisible()) {
 
+					updateSizeWithOvverides();
```
paw3lx: Do you think it is a good idea to put this method here? It will be called every time during rendering.

arshaw: Sorry for the delay @paw3lx. I think the best solution would be to call `_calcSize` from `reinitView`

paw3lx:
Ok, I dit it. Is everything ok now?

arshaw:
yes, looks good @paw3lx thank you. i will get this into the next release.

arshaw:
released in [v3.1.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.1.0). thanks!

PR #3421, opened 2016-11-13
Added Bengali support

codeshah:
I am not sure about the `allDayHtml`. So left it as it is. If you can explain, then I will correct it too.

### DIFF
--- a/locale/bn.js
+++ b/locale/bn.js
@@ -5,7 +5,7 @@ $.fullCalendar.locale("bn", {
 		day: "দিন",
 		list: "লিস্ট"
 	},
-	allDayHtml: "Todo<br/>el día",
+	allDayHtml: "সারাদিন",
 	eventLimitText: "আরও দেখুন",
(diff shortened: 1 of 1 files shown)


arshaw:
I was wondering about that. It's the text that goes here:
screen shot 2016-11-17 at 12 46 29 am

arshaw:
is that enough information for you to be able to provide a good translation for it?

codeshah:
Thank you for the clarification @arshaw 😃
I have updated translated that string. Please merge if everything seems okay to you.
My twitter username is "codeshah" 😉

arshaw:
thanks!

arshaw:
I have merged theses changes into the `next-release` branch, however, in order for translations to be correctly compiled, there needs to be a jquery-ui translations file available, and there is not:
https://github.com/jquery/jquery-ui/tree/master/ui/i18n

I eventually want to make fullcalendar translations independent from jquery-ui, but that's another project I'll undertake in the future.

PR #3426, opened 2016-11-17
update some translate

horkenw:
fix words translated error

### DIFF
--- a/locale/zh-tw.js
+++ b/locale/zh-tw.js
@@ -1,12 +1,14 @@
 
 $.fullCalendar.locale("zh-tw", {
 	buttonText: {
-		month: "月",
-		week: "週",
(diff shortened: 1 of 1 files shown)


arshaw:
Would you mind giving a little more info on why the current translations are bad, and why yours are better?

Also, it looks like your `eventLimitText` just returns a static string. It should ideally include the number of events that it was given, somewhere in the text.

Also, it looks like you committed spaces. The standard for these files is tabs.

horkenw:
sure, 
I make some explain about my change.
I was change some translate because the original translation was use simplified Chinese, so I just change it to traditional Chinese.
and some words were wrong used just according to location where we live, so I change made it more suitable

arshaw:
thanks

Issue #2907, opened 2015-10-21
eventOrder, precedence over date/duration


pedromendesws:
Hi,

I'd like to request a feature that is kind of what has been requested here https://code.google.com/p/fullcalendar/issues/detail?id=364, but not exactly.

eventOrder allows a custom sort when the start dates and durations are equal. What I wanted was to override the sort event in any situation.

For example if I have an event that has a duration of 2 days, but starts at 10AM I want it to appear before my 1 hour duration event that starts at 9AM. Example image in the attachs.

Actually I only see this being usefull in the month view. Is this possible?

![fullcalendarcustomsort](https://cloud.githubusercontent.com/assets/15229856/10634510/3d67d1ea-77ea-11e5-8e36-3ed619372111.png)

Thank you!

arshaw:
same request from @sboele:
> The current eventOrder funcionality: "Determines the vertical ordering events that have the same dates / times." according to the docs. It would be nice if you could change the order of the events by sorting by other properties first before sorting date/time. Now it only uses this option when events have the same date/time and duration. This way you can 'group' events together in a day view. For example: open and closing timewindows that are on the same day but you want them grouped together.

matiaslarsson:
Yes, please! Of course one can monkey patch compareEventSegs to, for example, remove the duration check but setting eventOrder specifically should really do what is says on the box. ;)
```
// A cmp function for determining which segments should take visual priority
compareEventSegs: function(seg1, seg2) {
	return seg1.eventStartMS - seg2.eventStartMS || // earlier events go first
		//seg2.eventDurationMS - seg1.eventDurationMS || // tie? longer events go first
		seg2.event.allDay - seg1.event.allDay || // tie? put all-day events first (booleans cast to 0/1)
		compareByFieldSpecs(seg1.event, seg2.event, this.view.eventOrderSpecs);
}
```

nafg:
Any chance of this being implemented soon?

rahulthewall:
Any updates on this?

rodogir:
Would really like to see this implemented as well.

harohl:
I would like to control the event stacking on the Scheduler timeline view.  Our schedule has standby and superseded schedules that I would like to display below the active/primary schedule.

Huzefa-lms:
To achieve sorting with custom event object parameter in existing code please replace **compareEventSegs** with the below function. I have used **eventOrder** as custom parameter in event object

compareEventSegs: function(t, e) {
[...]

PR #3446, opened 2016-12-07
#8479 add noSort option

louis-sanna:
Add option in case we want to show event in the exact order we added them.


arshaw:
Thanks for this quick solution. I'd rather wait until `eventOrder` undergoes these API changes:
https://github.com/fullcalendar/fullcalendar/issues/2907#issuecomment-315481115

Then, it would make sense to do `eventOrder:false` to achieve the behavior you desire.

Issue #3438, opened 2016-11-29
hide the heading row with columnFormat:false


Doc999tor:
AgendaDay view displays events for just one day. 
If a custom header is in use, the day title is unnecessary. 
I propose to add a view-specific `columnFormat: false` option which will empty the whole `thead.fc-head` or `td.fc-head-container`.

arshaw:
makes sense.

Doc999tor:
I'll send a pull request in a few days if I'll have a time.

Rouche:
Do you think this one could be taken care of at the same time? 
#3331

NickJorens:
Is there any progress? 
I'd love for this to be an option!

Rouche:
No idea. You can build locally with my pull request i guess.

@arshaw Wanted to change it for dom manipulation hook. Wich requires a lot more change. Too much in fact.
Because right now, ```renderHeadDateCellHtml``` is only generating plain html string.

Personally, making it possible to customise the Html generated is enough. For now...

caseyjhol:
@arshaw When `columnHead` is false, would you rather have an empty `<thead>` element, or do you want it gone completely?

arshaw:
sorry for the delay @caseyjhol . If `columnHead` is `false`, I'd expect the `<thead>` to be completely gone.

arshaw:
Released in [v3.6.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.6.0)

docs: https://fullcalendar.io/docs/display/columnHeader/

(renamed to *columnHeader*)

Issue #3331, opened 2016-09-06
rendering hooks for day/slot headings


smc395:
Was wondering if it is possible to add images/icons and change the CSS to calendar headings where (i.e. Sun 9/4) would be and in an event. Also if it was possible to tie actions to the heading like causing a panel to drop down.

Another question is when switching between month, week, and day views if animations would persist. So for example on the week view when I click on a day it brings a panel to drop down and then I click on the month view. If I click on the week view again will the panel still be dropped down? Thank you!

arshaw:
Would be cool to provide a rendering callback function as well as let the slot heading format text accept an HTML string somehow.

Rouche:
After reading the idea i changed the code of my temporary fix to this.
Works well for now, but ill wait for the real fix instead of PR.

```
	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
		var view = this.view;

		var headerText = typeof this.colHeadFormat === 'function' ? this.colHeadFormat(date, colspan) : htmlEscape(date.format(this.colHeadFormat));

		return '' +
			'<th class="fc-day-header ' + view.widgetHeaderClass + ' fc-' + dayIDs[date.day()] + '"' +
				(this.rowCnt === 1 ?
					' data-date="' + date.format('YYYY-MM-DD') + '"' :
					'') +
				(colspan > 1 ?
					' colspan="' + colspan + '"' :
					'') +
				(otherAttrs ?
					' ' + otherAttrs :
					'') +
				'>' +
				// don't make a link if the heading could represent multiple days, or if there's only one day (forceOff)
				view.buildGotoAnchorHtml(
					{ date: date, forceOff: this.rowCnt > 1 || this.colCnt === 1 },
					headerText // inner HTML
				) +
			'</th>';
	},

Rouche:
Would this solution be enough? Allow the ```columnFormat``` to be function? Or maybe theres a bigger picture i dont see?

Then in ```renderHeadDateCellHtml``` :
```var headerHtml = typeof this.colHeadFormat === 'function' ? this.colHeadFormat(date, colspan) : htmlEscape(date.format(this.colHeadFormat));```

Will PR if ok.

arshaw:
same issue: https://github.com/fullcalendar/fullcalendar/issues/3078

arshaw:
closing because this ticket if a duplicate of:
#3078
&uarr;
[...]

PR #3447, opened 2016-12-07
rendering hooks for day heading

### LINKED ISSUE #3078: expose rendering hooks for day headers
[requested by](https://github.com/fullcalendar/fullcalendar-scheduler/issues/142) @iamkoby and @mattdanna:

> Currently the only ability is to edit `slotLabelFormat`. 
> Let's say the format is "dd DD", I would love the option to edit css of that column but I can't separate classes of the "dd" and the "DD". The desired result is:
> screenshot 2016-02-24 12 03 02
> 
> I am now patching this with editing the html in the viewRender function:
> 
> ```
> viewRender: function(view, element){
>     element.find('.fc-head .fc-time-area th.fc-widget-header .fc-cell-text').each(function(i, col){
>         var content = $(col).text().trim().split(',');
>         $(col).html('<span class="day-name">' + content[0] + '</span><span class="day-num">' + content[1] + '</span>');
>     });
> },
> ```

Rouche:
Too bad i could not make it in time for 3.1.0

It should satisfy all the needs in every issues.

i was wondering if i should skip all the ```<tr>``` in ```renderHeadTrHtml``` when columnFormat is false

-----

this is a solution for the following issues: #3078, #3331, #3438

### DIFF
--- a/src/common/DayTableMixin.js
+++ b/src/common/DayTableMixin.js
@@ -67,8 +67,8 @@ var DayTableMixin = FC.DayTableMixin = {
 
 		this.colHeadFormat = this.view.opt('columnFormat');
 		if (this.colHeadFormat !== false) {
-		    this.colHeadFormat = this.colHeadFormat || this.computeColHeadFormat();
-	    }
+			this.colHeadFormat = this.colHeadFormat || this.computeColHeadFormat();
+		}
 	},
 
 
@@ -298,9 +298,9 @@ var DayTableMixin = FC.DayTableMixin = {
 	// (colspan should be no different)
 	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
 
-        if (this.colHeadFormat === false) {
-            return '';
-        }
+		if (this.colHeadFormat === false) {
+			return '';
+		}
 
 		var view = this.view;
 		var classNames = [
@@ -320,10 +320,10 @@ var DayTableMixin = FC.DayTableMixin = {
 			classNames.push('fc-' + dayIDs[date.day()]); // only add the day-of-week class
 		}
 
-        var headerHtml = typeof this.colHeadFormat === 'function' ? this.colHeadFormat(date, colspan, otherAttrs) : htmlEscape(date.format(this.colHeadFormat));
+		var headerHtml = typeof this.colHeadFormat === 'function' ? this.colHeadFormat(date, colspan, otherAttrs) : htmlEscape(date.format(this.colHeadFormat));
 
 		return '' +
-            '<th class="' + classNames.join(' ') + '"' +
+			'<th class="' + classNames.join(' ') + '"' +
 				(this.rowCnt === 1 ?
 					' data-date="' + date.format('YYYY-MM-DD') + '"' :
 					'') +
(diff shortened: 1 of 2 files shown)


arshaw:
thanks for the PR @Rouche 

i was actually hoping the "hook" would allow free-form DOM manipulation rather than just returning HTML. was hoping it could be similar to [eventRender](https://fullcalendar.io/docs/event_rendering/eventRender/)

lemme know if you disagree. let's discuss API more.

Rouche:
Well, how could we do this when a client just want to set the text/html inside the header.
There is a lot of infomation added that we need to keep on parent:
* fc-day-header
* view.widgetHeaderClass
* this.getDayClasses
* fc-' + dayIDs[date.day()]
* data-date
* colspan
Did i miss any?

Imo, no one should play with parent HTML, in eventRender we receive the element, and we do not modify it, we just add the rendering inside (its childs) Maybe we limit ourselves too much but its safer this way and less regression prone.

Rouche:
Biggest problem here is `renderHeadHtml > renderHeadTrHtml > renderHeadDateCellsHtml > renderHeadDateCellHtml` call chain dont manipulate dom, but html.

Thats why i need to reconvert with ```[0].outerHtml;``` and will probably remove dynamically added callbacks.

To do it clean, I would need to modify all the call chain to create HTML element from ```renderHeadHtml```

Rouche:
Updated for 3.3.0

src/common/DayTableMixin.js
```diff
@@ -293,6 +297,11 @@ var DayTableMixin = FC.DayTableMixin = {
 	// TODO: when internalApiVersion, accept an object for HTML attributes
 	// (colspan should be no different)
 	renderHeadDateCellHtml: function(date, colspan, otherAttrs) {
+
+		if (this.colHeadFormat === false) {
```
arshaw: if `false`, wouldn't we want to avoid rendering the whole `<tr>` rather than just the cells?

src/common/DayTableMixin.js
```diff
@@ -311,8 +320,10 @@ var DayTableMixin = FC.DayTableMixin = {
 			classNames.push('fc-' + dayIDs[date.day()]); // only add the day-of-week class
 		}
 
+		var headerHtml = typeof this.colHeadFormat === 'function' ? this.colHeadFormat(date, colspan, otherAttrs) : htmlEscape(date.format(this.colHeadFormat));
```
arshaw: rather than send HTML to the hook, which is clunkier to manipulate, i'd rather find a way to pass the jQuery DOM element to the hook. I know this will be awkward, as you previously stated. A technique you can use is to query the DOM element AFTER it's already created. Look it up via a className that you know will be there.

Rouche: Oh, thats an idea. Then the client would have the choice to empty the content and replace it with anything.

I will add a new option instead of reusing colHeadFormat, but i will keep the false option.

Ill work on this when i have a couple minutes.

src/common/DayTableMixin.js
Rouche: yes, it could be done directly in ```renderHeadHtml``` and return an empty string.

PR #3449, opened 2016-12-07
Use target property alongside url property with anchor tag

DanielLangdon:
It is great that the latest version of FullCalendar outputs items on the calendar as proper <a> tags.  I like the way it searches for a property called "url" on the data element and uses it to set the href attribute on the <a> tag.  However, in some applications, I need to be able to specify that the url should open in a new window/tab, so I propose that we look for a "target" property alongside the url element, and if it is found, use it to set the target attribute on the <a> tag.

### DIFF
--- a/src/common/DayGrid.events.js
+++ b/src/common/DayGrid.events.js
@@ -123,6 +123,10 @@ DayGrid.mixin({
 					' href="' + htmlEscape(event.url) + '"' :
 					''
 					) +
+                (event.url && event.target ?
+                    ' target="' + htmlEscape(event.target) + '"' :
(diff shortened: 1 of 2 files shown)


arshaw:
my current stance on requests like this is to move the logic to [eventRender](https://fullcalendar.io/docs/event_rendering/eventRender/) or [eventClick](https://fullcalendar.io/docs/mouse/eventClick/)

please see the source code of the [google calendar demo](https://fullcalendar.io/js/fullcalendar-3.1.0/demos/gcal.html). it hooks into `eventClick`.

Issue #2636, opened 2015-08-19
Duplicate week number in date range formatting text


arshaw:
Originally reported on Google Code with ID 2371

```
When you use 'W' in titleFormat properties and firstDay set to 1, in agendaWeek view
you display only one week. I expect the title display only the current week number,
but it display the number twice.

See in jsfiddle : http://jsfiddle.net/3E8nk/879/

I can propose this fix, but I'm a noob in fullCalendar and not very experimented in
jquery.

In formatRangeWithChuncks function : 

function formatRangeWithChunks(date1, date2, chunks, separator, isRTL) {
    var chunkStr; // the rendering of the chunk
    var leftI;
    var leftStr = '';
    var rightI;
    var rightStr = '';
    var middleI;
    var middleStr1 = '';
    var middleStr2 = '';
    var middleStr = '';

        //.... skip some code

    if (middleStr1 || middleStr2) {
        if (isRTL) {
            middleStr = middleStr2 + separator + middleStr1;
        }
        else {
            middleStr = middleStr1 + separator + middleStr2;
        }
    }

        //My Fix
    if (middleStr1 == middleStr2) {
            middleStr = middleStr1;
    }
        //End of fix

    return leftStr + middleStr + rightStr;
}

```

Reported by `hydreliox` on 2014-11-21 09:44:46

Imported with **1** stars.

arshaw:
this won't be a problem in v4/v5 because date formatting is done with the built-in Intl API

PR #3467, opened 2016-12-20
add week to similarUnitMap (also fixes issue 2371)

runemoennike:
I humbly request that this change be merged into master. It fixes a problem I'm encountering where week numbers are duplicated in the title, when the start and end weeks for a view are equal.

If not, can you guys elaborate on why it should not be merged? I can see the bug ( https://github.com/fullcalendar/fullcalendar/issues/2636 ) has been open for quite a while now, and it's marked as "todo" in the code.

### DIFF
--- a/src/date-formatting.js
+++ b/src/date-formatting.js
@@ -150,6 +150,8 @@ function formatRangeWithChunks(date1, date2, chunks, separator, isRTL) {
 var similarUnitMap = {
 	Y: 'year',
 	M: 'month',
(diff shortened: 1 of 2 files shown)


arshaw:
thank you for making this PR and for writing a test. I will try my best to get this into the next release.

runemoennike:
Just wanted to point out that @r-springer made the all the code and the test for this PR, for good measure :)

Issue #1813, opened 2015-08-19
Jalaali date system support


arshaw:
Originally reported on Google Code with ID 1546

```
Hi
i have to build a jalali calender and every thing was ok expect one problem
month days are different from gorgeous

for example second month in gorgeous(feb) have 28 days but second month in jalali has
31 days!

the days count are look like this 
31-31-31-31-31-31-30-30-30-30-30-29 (30 every 4 years)

Thanks
```

Reported by `unireseller` on 2012-09-23 11:28:21

Imported with **9** stars.

ehsandanesh:
Hi everybody I added Jalaali calendar to version 2.6.0 and solved previous issues in this update you can find it from here:
https://github.com/ehsandanesh/fullcalendar

Natico:
Hi , in last week I changed & added Jalaali calendar to FullCalendar version 30.1 .
you can find it from here:
[https://github.com/Natico/fullcalendar-Jalaali](https://github.com/Natico/fullcalendar-Jalaali)

espen:
Can this be made into a pull request @Natico ?

mamos98:
@Natico Naser Jan Can you please send a Pull request to make your works available for all the users who want Persian Calendar Support maybe? 👍 
Thanks in Advance!

arshaw:
It seems like this would be difficult to make into a pull request without littering "jalaali" all throughout the codebase. I'd like to think of the Jalaali calendar system as just one system of many, similar to a timezone or a locale. As such, momentjs should handle this rather than special-cases in fullcalendar.

momentjs is discussing this concept here: https://github.com/moment/moment/issues/1454
(I'd encourage you to upvote that ticket)

I imagine the API will turn out something like this:

```
var m = moment(str, { calendar: 'jalaali' });
m.calendar(); // getter, "jalaali"
m.calendar('hijiri'); // setter
```

The important concept is that the *moment* will store the state about what calendar system it is using. The owner of the moment object won't need to store that state separately themselves.

If things progress in that direction, then the same `.day()`, `.week()`, `.month()`, `.year()` methods could be used. And there would be easily pluggable formatting/parsing hooks. And if that were the case, the [moment-jalaali](https://github.com/jalaali/moment-jalaali) extension could be much simpler that it is, free of all that duplicate core momentjs code.

[...]

PR #3532, opened 2017-02-16
support Buddhist year format string

nattster:
support Buddhist year format string by adding `B` to `specialTokens`.

need to add `B` to largeTokenMap to make formatRange works properly.

### DIFF
--- a/src/date-formatting.js
+++ b/src/date-formatting.js
@@ -39,6 +39,9 @@ var specialTokens = {
 	},
 	T: function(date) { // "A" or "P"
 		return oldMomentFormat(date, 'A').charAt(0);
+	},
+	B: function(date) {
+		return date.year() + 543;
 	}
 };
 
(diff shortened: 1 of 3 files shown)


synycboom:
:+1:

arshaw:
Thanks! Ignorant question: are there more aspects of the Buddhist calendar you'd want supported by FullCalendar other than just year?

nattster:
Moment.js's locale handles most of the requirement we need now (e.g. Thai month and day name).

Buddhist month was used more in astrology, which is quite uncommon in normal web application.
macOS localized only the year part:

be_year

So I think that's would be all for Buddhist calendar. Thank you!

arshaw:
@nattster, after thinking about this more, it is my opinion that other [non-Gregorian] calendar systems should not be special-cased into fullcalendar's code. See me thoughts on the Jalaali thread: https://github.com/fullcalendar/fullcalendar/issues/1813#issuecomment-281620201

So, I don't like creating a new `B` formatting token. I'd much rather expose `largeTokenMap` and `specialTokens` and have the caller of FullCalendar's API monkeypatch support for `B` in there.

Could you expose those?
```
$.fullCalendar.moment.specialTokens.B = // whatever
$.fullCalendar.moment.largeTokenMap.B = // whatever
```

You can keep your test essentially the same, but just do the monkeypatch code before the meat of the tests run. Please also remember to clean up the monkeypatched properties, because they are global state that runs across all tests:

```
delete $.fullCalendar.moment.specialTokens.B;
delete $.fullCalendar.moment.largeTokenMap.B;
```

Does that make sense?

arshaw:
closing this issue because the v4/v5 codebase is no longer based on moment, but please see the discussion in #5576

Issue #3090, opened 2016-03-14
agenda event overlap consider computed width


lorinkoz:
As of now, when two or more events overlap in time, fullcalendar calculates the event width based on how many events are on parallel, using percentage widths. I'm doing a project where I must display a generic "availability" event along with actual events, so that a scheduler person may know when someone is available. I was trying to make the availability event 10px width, so that it doesn't interfere with actual events, but then, even if I managed to set the width of the target events, the overlapping events do not adjust the left property based on that and are displayed as if the former event was full width. Background events are not working for me here, as I have to move and resize availability.

![example](https://cloud.githubusercontent.com/assets/3628106/13744696/d2abdbd8-e9bf-11e5-94b8-1d3d07f0c6d4.png)

arshaw:
will `slotEventOverlap:false` help you?
http://fullcalendar.io/docs/agenda/slotEventOverlap/

lorinkoz:
Nope, I already have `slotEventOVerlap:false`, I just want to be able to set the width of events or eventSources and have the overlapping events adjust their `left` according to the width of the first event. In the picture, the event 8am-8.30am is still displayed with `left:50%` despite the overlapping event is not using 50% of the width.

arshaw:
I see. How would you ideally like to set that 10px width? via CSS or through JS?

I'll keep this issue open in case others have the same need. Somewhat similar to #961 in that the event positioning internals need to be aware of compute dimensions.

lorinkoz:
Ideally, this would be configured in an event or eventSource basis, including some `width` property in the object, such as:

    {
        ...
        width: '10px', // or '10%'
        ...
    }

Thank you so much for considering this :)

lorinkoz:
Is there any way to contact you to discuss several tweaks for a custom version?

lorinkoz:
Nevermind, someone posted an answer on my stackoverflow question that seems to be working as of now. https://stackoverflow.com/questions/35579749/increase-clickable-area-of-slots-behind-fullcalendar-event

Issue #961, opened 2015-08-19
Min-height for timeGrid events. Render events by computed height


arshaw:
Originally reported on Google Code with ID 691

```
Can you add a "Minimum event height" option to stop the events becoming too thin to
read?  I've got events in my calendar that are only 5 minutes long and they appear
as 4 pixel tall boxes with no text in them.  It would be great if they stopped shrinking
at about 15 pixels so the title was still visible.

Thanks

Russell
```

Reported by `russell.godsmark` on 2010-10-29 10:30:38

Imported with **15** stars.

fsboehme:
+1 – this is what my day view looks like in some cases where events are short

![league___tournament_scheduler___leaguelobster___and_google_hangouts_-_fsboehme_gmail_com](https://cloud.githubusercontent.com/assets/4282394/19484550/fe94a268-9525-11e6-9108-0082edef2f8a.jpg)

OWeiha:
+1

arshaw:
Released in [v3.6.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.6.0)

docs: https://fullcalendar.io/docs/agenda/agendaEventMinHeight/

arshaw:
I incorrectly close this ticket. The new `agendaEventHeight` setting provides a workaround, but the end-goal is to have the computed height serve as the min-height.

Spyridox:
`agendaEventMinHeight` works for changing event height, but how is it possible to make the cells themselves taller? Changing css properties does change the cells, but then the events are not moved where they should be.

danielok1993:
Any update/tips on the cell height issue? The `agendaEventHeight` is great, but due to cells not growing larger, it looks wrong. Thanks

ppazos:
This is not only for the agenta view, also in the timeGridWeek when there are a more than 2 events with small durations together it's impossible to read the event's text.

Would be desirable to:

1. allow setting the time row height by configuration
2. allow hiding certain hours that are not used in the timeGridWeek and agenda views

It seems both things can't be done right now in v5

thegiantbeast:
@arshaw I'm having the same problem for short events and tried the `agendaEventHeight` but it seems it was removed on v5, then I looked into using `min-height` for `.fc-media-screen .fc-timegrid-event-harness` selector, which was described in this thread too, and although it kinda works, the events that overlap start to be hidden behind each other, so it seems the library doesn't take the computed styles into consideration when rendering the events.

[...]

PR #3534, opened 2017-02-16
Enabled setting of minSegHeight

francescoGiargoni:
Enables to set minSegHeight in view options, in order to customize the
minimum height causing the event to split text in 2 lines (one for the
time, the other for the content)

### DIFF
--- a/src/common/TimeGrid.events.js
+++ b/src/common/TimeGrid.events.js
@@ -468,8 +468,8 @@ TimeGrid.mixin({
 			seg = segs[i];
 			seg.el.css(this.generateFgSegHorizontalCss(seg));
 
-			// if the height is short, add a className for alternate styling
-			if (seg.bottom - seg.top < 30) {
+			// if the height is short (or shorter than set in view options), add a className for alternate styling
+			if (seg.bottom - seg.top < (this.view.opt('minSegHeight') || 15)*2 ) {
 				seg.el.addClass('fc-short');
 			}
 		}


arshaw:
Hi @francescoGiargoni , thanks for the PR.

I originally coded that 30px value as a hack. The *ideal* behavior is to detect when the event element is short enough where its title is obscured, and only then consider the element to be "short" and push its title into the top. This ideal behavior involves querying the DOM for these coordinates after the element has already been partially rendered. This is somewhat difficult with the way the code is set up. At the point in the code where `fc-short` is applied, the element is not in the DOM yet.

There are a number of other tickets that require this premise:
https://github.com/fullcalendar/fullcalendar/issues/961
https://github.com/fullcalendar/fullcalendar/issues/2853
https://github.com/fullcalendar/fullcalendar/issues/3090

I'd prefer to get this working the "right way" than add another setting. I'll consider this PR "blocked".

francescoGiargoni:
Got it, I'll drop by if I happen to find a way to achieve what you stated.

arshaw:
I ended up not implementing this as `minSegHeight`like you wanted, but rather has a more internal setting that I expose through `FullCalendar.config.timeGridEventCondensedHeight`. See: https://github.com/fullcalendar/fullcalendar/blob/v5.1.0/packages/timegrid/src/TimeCol.tsx#L32

I don't plan on introducing a fully-public setting because I plan to have this computed automatically in the future (based on whether the bottom event bound goes about the event title's bottom bound)

Issue #429, opened 2015-08-19
Restrict date range


arshaw:
Originally reported on Google Code with ID 158

```
Is there a way to restrict the viewable dates, i.e. next 2 years only?
```

Reported by `naterowe82` on 2009-10-30 15:12:56

Imported with **31** stars.

arca223:
Hi arshaw !

Any news about this to be released ?
Is it going to be only the navigation restricted or also a way to disable click if the date range < whole month or whole week ?

Something interesting would be to disable cells not in the range, and disable the prev/next when reaching the edges. If event is on 2 weeks, you got the navigation up

![fullcalendarft](https://cloud.githubusercontent.com/assets/9930499/14598571/30483a26-0553-11e6-879c-a5ffabed14d1.png)

I hope it was the right place to post, I don't know if I had to remake an issue since you'd merge it here

Thanks for your help. (PS : sry for the picture fastly made on paint)

arshaw:
Released in **v3.3.0**
Blog post: https://fullcalendar.io/blog/2017/03/fullcalendar-33-and-scheduler-16/

[validRange](https://fullcalendar.io/docs/current_date/validRange/)

Bridgeway:
How do you set the range when the architecture is as follows:

this.eventFingerprint = function (e) {
                if (!e._id) {
                    e._id = eventSerialId++;
                }

                var extraSignature = extraEventSignature({
                    event : e
                }) || '';
                var start = moment.isMoment(e.start) ? e.start.unix() : (e.start ? moment(e.start).unix() : '');
                var end = moment.isMoment(e.end) ? e.end.unix() : (e.end ? moment(e.end).unix() : '');

                // This extracts all the information we need from the event. http://jsperf.com/angular-calendar-events-fingerprint/3
                return [e._id, e.id || '', e.title || '', e.url || '', start, end, e.allDay || '', e.className || '', extraSignature].join('');
            };

acerix:
@Bridgeway Please refer to the [support page](https://fullcalendar.io/support/) and use [Stack Overflow](https://stackoverflow.com/questions/tagged/fullcalendar) for help.  If this is a bug, please supply a [runnable, stripped-down demonstration](http://fullcalendar.io/wiki/Reduced-Test-Cases/).

PR #3539, opened 2017-02-17
limit next and prev clicks

yudeikisorta:
Added 2 new properties to calendar: **limitPrevClicks** and **limitNextClicks** with integer values in order to limit Prev and Next navigation according to these parameters

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -508,6 +508,7 @@ function Calendar_constructor(element, overrides) {
 	var windowResizeProxy; // wraps the windowResize function
 	var ignoreWindowResize = 0;
 	var date; // unzoned
+	var navigationButtonClicks = 0;
 
 
 
@@ -680,6 +681,7 @@ function Calendar_constructor(element, overrides) {
 					// need to do this after View::render, so dates are calculated
 					// NOTE: view updates title text proactively
 					updateToolbarsTodayButton();
+					updateNavigationButtons();
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/Calendar.js
```diff
@@ -883,6 +885,31 @@ function Calendar_constructor(element, overrides) {
 	}
 
 
+	function updateNavigationButtons() {
+	    var nextButtonLimit = parseInt(t.options.limitNextClicks, 10),
+            prevButtonLimit = parseInt(t.options.limitPrevClicks, 10);
```
criztovyl: I think the default should be that the option is disabled, not limit to +-10.
As it's numeric I would suggest `-1`, meaning the limit is disabled.

src/Calendar.js
```diff
@@ -883,6 +885,31 @@ function Calendar_constructor(element, overrides) {
...
+
+		//checking next button
+        if (nextButtonLimit >= 0) {
```
criztovyl: You will need to check for `-1` here first. If `-1`, should be no-op.

src/Calendar.js
```diff
@@ -883,6 +885,31 @@ function Calendar_constructor(element, overrides) {
...
+            if (navigationButtonClicks >= nextButtonLimit) {
+                toolbarsManager.proxyCall('disableButton', 'next');
+            } else {
+                toolbarsManager.proxyCall('enableButton', 'next');
+            }
+        }
+
+        //checking prev button
+        if (prevButtonLimit >= 0) {
```
criztovyl: Same as above.

src/Calendar.js
yudeikisorta: 10 is not the default option, is the redix parameter for parseInt function 

https://www.w3schools.com/jsref/jsref_parseint.asp
The radix parameter is used to specify which numeral system to be used, for example, a radix of 16 (hexadecimal) indicates that the number in the string should be parsed from a hexadecimal number to a decimal number

by default the option is disabled unless its value is > -1

criztovyl: I'm sorry, I missed the radix ._.

But how is it disabled when it's below 0?

arshaw:
thanks @yudeikisorta , this is a nice hack, but i'd prefer to limit the allowed navigatable date range via explicit start/end dates or forward/backward date delta, rather than rely on "clicks" which might equate to a different jumped duration depending on the view.

more discussion here: https://github.com/fullcalendar/fullcalendar/issues/429

arshaw:
i mentioned previously that I'd rather a date range limit the user's ability to navigate rather than a number of clicks.

It is now possible to restrict the date range: https://fullcalendar.io/docs/current_date/validRange/

So I will close this PR

PR #3557, opened 2017-02-27
Fix date range output

jleja:
Fixes date output for start/end range in Agenda mode by not outputting `undefined` when a value doesn't exist.


arshaw:
is there a corresponding bug ticket for this?

also, it looks like you committed the built dist files. if you post a follow-up PR, please omit that.

jleja:
Hi Adam, sorry about that, thought I had deleted everything.  At first I thought there was a bug when I updated the FC component in our project, but it turns out it was getting loaded twice, and causing some weird issues (hence me committing the build file to try something out).  It's since been resolved.

arshaw:
k thanks for letting me know. i will keep this closed.

Issue #2334, opened 2015-08-19
Twitter Bootstrap support


arshaw:
Originally reported on Google Code with ID 2069

```
Hi,

Twitter Bootstrap is more and more popular, so could you please add the ability to
use Bootstrap's classes instead of yours ?

With this option it will be easier to implement FullCalendar in an existing (bootstrap)
project. 
Also using default classes will allow people to use all themes already available (from
http://bootswatch.com/ , https://wrapbootstrap.com/?ref=bsw or any others)...

I'm not a CSS/jQuery/Bootstrap ninja but you probably can do it by defining all your
classes in a specific files that must be included before the main script. One files
will use Bootstrap classes, an other one will use jQuery UI, etc...

My english is really poor so I hope all of this is readable !

Thanks !
```

Reported by `schmittounet` on 2014-01-13 14:12:10

Imported with **6** stars.

arshaw:
this is queued for the next release. see https://github.com/fullcalendar/fullcalendar/pull/3566. closing

arshaw:
Released in v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/#more-257))

Issue #3564, opened 2017-03-01
Add a CSS file to dist with the minimal styles needed for themes


joankaradimov:
When instantiating a fullcalendar with a `theme: true` option only a fraction of the styles in `fullcalendar.css` is used.

Would it possible to build an additional CSS file (e.g. `fullcalendar.external-theme.css`) that only includes the minimal amount of styles needed to get fullcalendar running with an external theme?

I know it's not a huge gain (with non-minified/non-gzipped `fullcalendar.css` being only 31KB). But I am assuming that the separate CSS files that go into `fullcalendar.css` already have that separation to some degree. So an implementation of this could be as simple as adding:
```json
"fullcalendar.external-theme.css": [
  "main.css"
]
```
... to `src.json`. If that is not the case then maybe this feature request can be ignored.

arshaw:
i agree, this would be nice.

if/when there's a "theme system" refactor, like this:
https://github.com/fullcalendar/fullcalendar/issues/2334

...that will be a good time to reevaluate.

arshaw:
@joankaradimov I plan to refactor nearly all of FullCalendar's CSS while tackling this issue: #2853 . I plan to make things BEM-y (which will minimize rules fighting with each other for precedence). I think that will be the best time to split up the skeleton CSS from the "standard theme" CSS more effectively.

arshaw:
This will be implemented for v7. More info: https://github.com/fullcalendar/fullcalendar/issues/7907

Issue #2836, opened 2015-08-24
Full Calendar Material Design Theme


jackyliang:
Hey all.

I made a Material Design-like theme for Full Calendar which I've opened sourced under my [repo](https://github.com/jackyliang/Material-Design-For-Full-Calendar)

<img src="http://i.imgur.com/TH3VsJU.gif">

What the actual Android Material Design calendar looks like:
<img src="http://dab1nmslvvntp.cloudfront.net/wp-content/uploads/2014/09/1410153384GIF2.gif">

While FullCalendar is pretty awesome, it looks pretty terrible without any styling. I was inspired by Google's Material Design, while coincidentally I needed a fitting theme for my Drexel [Schedulizer](http://loop.tf/schedulizer) service. Doing some basic research, it didn't seem like there were any Material Design calendar plugins or themes that were as flexible as FullCalendar's functionality, so I took matters to my own hands and put this together. 

Currently it only supports the `agendaWeekly` view since that was my context of use. Feel free to contribute if you desire more usages.

StudioJunkyard:
Nice, I believe Bootstrap v4, is implementing a few Material type components as well.

jackyliang:
@StudioJunkyard Sweet. If you can't wait, [this](http://fezvrasta.github.io/bootstrap-material-design/bootstrap-elements.html) is what I currently use for Bootstrap Material Design. I would say the results are fairly good. Angular.js itself also has Google's official Material theme which is incredibly done.

StudioJunkyard:
+1

dtmonterrey:
Very nice...

tomvlk:
:+1:

piyushchauhan2011:
:+1:

kanpinar:
+5

espen:
Looks great. And it's really nice to see community addons to Fullcalendar. Perhaps there should be a section in the readme with links to this and other plugins/themes @arshaw ? Would probably encourage more open source addons instead of private development like I'm sure many have done to customize Fullcalendar.

jackyliang:
@espen Love this idea! Finding other themes for FullCalendar was very hard, if not non-existent. I think adding this theme alongside other private/public themes will spur development more!

arshaw:
This looks reaaally nice.

You were able to achieve this [entirely through CSS](https://github.com/jackyliang/Material-Design-For-Full-Calendar/blob/master/_materialFullCalendar.scss) it looks like? Even that day-expand animation?

[...]

Issue #1436, opened 2015-08-19
Add jQuery UI Theme Switcher to the demo page


arshaw:
Originally reported on Google Code with ID 1169

```
Might be useful to add the dynamic theme switcher to the query ui theme demo page.

In the header add:
<link href="//ajax.googleapis.com/ajax/libs/jqueryui/1.8.16/themes/ui-lightness/jquery-ui.css"
rel="stylesheet" type="text/css"/>

...add the Switch placement code:
<div id="switcher"></div>

... and then add this to your JS
$('#switcher').themeswitcher();

That's it (other than you can add some positioning CSS for the switcher div)
```

Reported by `StlDoug` on 2011-11-06 19:49:13

Imported with **1** stars.

arshaw:
this is queued for next release. closing. there will be a demo html file that lets you preview multiple jquery ui themes as well as Bootstrap themes. see https://github.com/fullcalendar/fullcalendar/pull/3566

arshaw:
Released in v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/))

https://fullcalendar.io/js/fullcalendar-3.5.0/demos/themes.html

PR #3566, opened 2017-03-02
Bootstrap themes

joankaradimov:
I've hacked together a simple implementation of #2334. It has some visual issues, but it's usable overall.

If this is still on the roadmap I could spend some time improving the implementation.

### DIFF
--- a/src/Calendar.render.js
+++ b/src/Calendar.render.js
@@ -51,6 +51,7 @@ Calendar.mixin({
 		this.optionsModel.watch('applyingThemeClasses', [ '?theme' ], function(opts) {
 			el.toggleClass('ui-widget', opts.theme);
 			el.toggleClass('fc-unthemed', !opts.theme);
+			_this.theme.setTheme(opts.theme);
 		});
 
 		// called immediately, and upon option change.
--- a/src/common/Theme.js
+++ b/src/common/Theme.js
@@ -18,6 +18,13 @@ var themes = {
 			buttonGroup: 'fc-button-group',
 			tableHeader: 'fc-widget-header',
 			tableContent: 'fc-widget-content'
+		},
+		iconClasses: {
+			close: 'fc-icon fc-icon-x',
+			prev: 'fc-icon fc-icon-left-single-arrow',
+			next: 'fc-icon fc-icon-right-single-arrow',
+			prevYear: 'fc-icon fc-icon-left-double-arrow',
+			nextYear: 'fc-icon fc-icon-right-double-arrow'
 		}
 	},
 	jQueryUI: {
@@ -39,6 +46,13 @@ var themes = {
 			buttonGroup: 'fc-button-group',
 			tableHeader: 'ui-widget-header',
 			tableContent: 'ui-widget-content'
+		},
+		iconClasses: {
+			close: 'ui-icon ui-icon-closethick',
+			prev: 'ui-icon ui-icon-circle-triangle-w',
+			next: 'ui-icon ui-icon-circle-triangle-e',
+			prevYear: 'ui-icon ui-icon-seek-prev',
+			nextYear: 'ui-icon ui-icon-seek-next'
 		}
 	},
 	bootstrap3: {
@@ -55,6 +69,13 @@ var themes = {
 			tableContent: 'panel-default',
 			tableGrid: 'table-bordered',
 			tableList: 'table'
+		},
+		iconClasses: {
+			close: 'fc-icon fc-icon-x',
+			prev: 'fc-icon fc-icon-left-single-arrow',
+			next: 'fc-icon fc-icon-right-single-arrow',
+			prevYear: 'fc-icon fc-icon-left-double-arrow',
+			nextYear: 'fc-icon fc-icon-right-double-arrow'
 		}
 	}
 };
@@ -76,5 +97,31 @@ FC.Theme = Class.extend({
 
 	getClass: function(key) {
 		return themes[this.theme].classes[key] || '';
+	},
+
+	getIconClass: function(buttonName) {
+		return themes[this.theme].iconClasses[buttonName] || '';
+	},
+
+	getIconClassWithOverride: function(buttonName, customButtonProps, calendar) {
+		if (this.theme === 'builtin') {
+			if (customButtonProps) {
+				return customButtonProps.icon;
+			} else if (!calendar.opt('buttonIcons')) {
+				return undefined;
+			} else if (calendar.opt('buttonIcons')[buttonName]) {
+				return 'fc-icon fc-icon-' + calendar.opt('buttonIcons')[buttonName];
+			}
+		} else if (this.theme === 'jQueryUI') {
+			if (customButtonProps) {
+				return customButtonProps.themeIcon;
+			} else if (!calendar.opt('themeButtonIcons')) {
+				return undefined;
+			} else if (calendar.opt('themeButtonIcons')[buttonName]) {
+				return 'ui-icon ui-icon-' + calendar.opt('themeButtonIcons')[buttonName];
+			}
+		}
+
+		return this.getIconClass(buttonName);
 	}
 });
--- a/src/common/Theme.js
+++ b/src/common/Theme.js
@@ -26,6 +26,14 @@ var themes = {
 		cornerLeft: 'ui-corner-left',
 		cornerRight: 'ui-corner-right',
 		buttonGroup: 'fc-button-group'
+	},
+	bootstrap3: {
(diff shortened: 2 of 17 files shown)


arshaw:
@joankaradimov this is really awesome! Can I offer some suggestions on how to make the code better, so that I can include it in a release?

All over the place there are statements like this:
```
FC.ThemeClasses.stateDown(tm)
```

It'd be great if we could rely on a method more closely related to the current object so that we don't need to query for the `tm` value each time. So:

```
// from within the calendar class
this.theme.stateDown();

// from within the view class
this.calendar.theme.stateDown();

// from within a Grid class
this.view.calendar.theme.stateDown()
```

So, you'd need to create a `theme` object that becomes a property of the calendar. The Calendar would be responsible for querying for the `theme` option ahead of time.

Make sense?

joankaradimov:
> Make sense?

It makes perfect sense. I hadn't realized that a calendar instance is always accessible from the scope of all objects that use themes.

joankaradimov:
Done. Thanks for the feedback. Tell me if/when you want me to squash.

One thing I was thinking about - I could iterate the `themes` hash and dynamically generate the functions for html classes in `Theme.js`. It would be shorter but less explicit. Any preferences on that one?

Any other feedback?

arshaw:
hi @joankaradimov , sorry for the delay. I prefer `this.calendar.theme.getClass('widgetHeader')` instead of `this.calendar.theme.widgetHeader()` . If you don't have time to make the changes, I can.

joankaradimov:
No problem. It's a quick change. I'll do it.

joankaradimov:
Done. The last test failure is the same as the other recent PRs and shouldn't happen when the tests are ran locally.

src/common/Theme.js
```diff
@@ -0,0 +1,58 @@
...
+
+	setTheme: function(theme) {
+		if (theme === true || theme === 'jQueryUI') {
+			this.theme = 'jQueryUI';
+		} else if (theme === 'bootstrap3') {
+			this.theme = 'bootstrap3';
+		} else {
+			this.theme = 'builtin';
+		}
+	},
```
criztovyl: I don't understand what the `if...else if...else` does, why hardcode possible themes there?

joankaradimov: The intent was to keep compatibility with the old `true`/`false`/`undefined` parameter. But I see why it's not obvious what this code does. I've updated the pull request.

arshaw:
Hi @joankaradimov , I was about to merge this into the upcoming release myself (was going to resolve the inevitable merge conflicts too), but decided to hold off because there are some QA issues:

- relies on jQuery UI button icons. i would rather not need people to include jqueryui+bootstrap theme CSS on the same page.
- in the demo it would really really nice to allow the user to select a theme manually, rather than relying on the random button
- harsh dark lines in agenda view: screen shot 2017-04-26 at 3 55 41 pm
- popover is broken: screen shot 2017-04-26 at 4 01 58 pm

src/common/Theme.js
```diff
@@ -0,0 +1,58 @@
+var themes = {
+    builtin: {
```
arshaw: mixing tabs and spaces

joankaradimov: I'll fix and force push this particular piece code.

joankaradimov:
> was going to resolve the inevitable merge conflicts too

I can do the rebase. I am not doing it now, though. I'll need some time to get it right.

> relies on jQuery UI button icons. i would rather not need people to include jqueryui+bootstrap theme CSS on the same page.

Using Glyphicons would be the easiest solution here. But I'm not fond of it as some people disable Glyphicons and use alternative solutions (me being one of those people). Let me think about it.

> in the demo it would really really nice to allow the user to select a theme manually, rather than relying on the random button

Done.

> harsh dark lines in agenda view: 

Done.

> popover is broken

I am yet to look at this one.

arshaw:
sounds great @joankaradimov, let me know if you have any questions.

joankaradimov:
I squashed, rebased, fixed some of the issues, and force pushed. Maybe a new review is in order.

What remains to be done is to figure out what to do with the button icons.

src/Calendar.render.js
```diff
@@ -51,6 +51,7 @@ Calendar.mixin({
 		this.optionsModel.watch('applyingThemeClasses', [ '?theme' ], function(opts) {
 			el.toggleClass('ui-widget', opts.theme);
 			el.toggleClass('fc-unthemed', !opts.theme);
+			_this.theme.setTheme(opts.theme);
```
joankaradimov: I am not sure this is the correct way to do this now. Is there a cleverer solution here?

joankaradimov:
I've updated the pull request.

The button icons are now part of the theme. If you pick the builtin theme or the jQuery UI theme you can manipulate the icons as before (via the [`buttonIcons`](https://fullcalendar.io/docs/display/buttonIcons/) and [`themeButtonIcons`](https://fullcalendar.io/docs/display/themeButtonIcons/) settings accordingly). The `icon` and `themeIcon` properties of [`customButtons`](https://fullcalendar.io/docs/display/customButtons/) should also work as before. Setting these will override the default icons.

Currently it is not possible to manipulate the icons of the new bootstrap theme. It should be fairly straightforward to expose an interface that allows the manipulation of the icons (maybe extending existing themes or creating new ones).

However, such a mechanism would be redundant with the backwards compatible one. So I've left it out for now. A suggestion on how to go with this would be welcome. If it's up to me I would deprecate the old way of setting icons and keep it for a while along a new way.

I've also created two separate bootstrap themes - one that uses glyphicons and one that does not.

arshaw:
hi @joankaradimov, I did some additional work on top your branch, the first part of which can be seen [in these commits](https://github.com/fullcalendar/fullcalendar/commits/8edbe9c252fe0bd625ba57484c701898dff70fde).

- Made each theme system more plugin-esque
- rnamed the theme `bootstrap3`
- new setting called `bootstrapGlyphicons` which can customize individual icons
- if you don't want to use glyphicons, set `bootstrapGlyphicons` to `false`. I think this achieves what you want, though it does not default the icons to the built-in theme's icons
- rename `built-in` theme to `standard`

I then merged all this into [the master branch](https://github.com/fullcalendar/fullcalendar/commits/master), which contains a big refactor of other misc stuff.

After that, the biggest thing I did was redo the demo page. It combines jQuery UI and Bootstrap into one demo file. Plz take a look (`demos/themes.html`).

This PR is now closed because it has been merged into master, but it still needs to be released. There's a bit more refactoring of other subsystems before the next release. If you want to make further changes, please base PRs against `master`.

joankaradimov:
Thanks!

> If you want to make further changes, please base PRs against master.

The only thing I'd like to suggest is that the color of the "today" cell is tweaked in the Bootstrap theme. I made this PR for it: https://github.com/fullcalendar/fullcalendar/pull/3745.

Other than that - it looks like these two issues can be closed now: 
 - https://github.com/fullcalendar/fullcalendar/issues/2334
 - https://github.com/fullcalendar/fullcalendar/issues/1436

Also the author of the Material Design Theme in https://github.com/fullcalendar/fullcalendar/issues/2836 might find it much easier to integrate his changes into master now.

And finally - I opened https://github.com/fullcalendar/fullcalendar/issues/3564 a while ago. I'll try to figure out what the separation of `common.css` into several files means for the issue. If there's only 5-6 KB to be gained, it should not be considered at all. But if more CSS can be split in a similar fashion, I'll be interested in making the pull requests.

arshaw:
@joankaradimov thanks for all the hard work coding as well as helping me keep track of these issues.

I've merged the "today" PR and commented on all the other ones you mentioned.

joankaradimov:
@arshaw Thanks for the cooperation and the guidance. And for the great project, of course!

I'm looking forward to the next release.

arshaw:
Released in v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/))

thanks @joankaradimov 

If you're curious, the delay in releasing after this was implemented was due to me trying to do continued refactoring after the merge. However, the refactoring became much more extensive than planned, so I decided to release the code at the time of the Bootstrap 3 merge, which was stable. I'm continuing to work on the refactor. More info on the blog.

joankaradimov:
> If you're curious, the delay in releasing after this was implemented was due to me trying to do continued refactoring after the merge

Yes, I've followed the commits. I'd love to see this progress so we can see more features and performance improvements in the future.

And considering the blog post - Bootstrap 4 support sounds great. I'd love to implement it once we have a final version (and hopefully support for the Bootswatch themes).

PR #3614, opened 2017-03-31
Change cursor for event

idoo:


espen:
Event should already have a pointer cursor? Or am I missing something here?

Issue #2999, opened 2016-01-12
Official version of FullCalendar via Packagist


marclaporte:
Hi!

There are many FullCalendar sources here:
https://packagist.org/search/?q=fullcalendar

Would be nice to have the official one.

Thanks!

espen:
Why? It's a JavaScript lib, not PHP.

arshaw:
Packagist seems like a Composer package directory, and Composer is PHP package, so I'm very hesitant to go down this route. However, I have seem other front-end projects (just momentjs i guess) define a `composer.json`

marclaporte:
It's common to have packages that have JavaScript but no PHP. This represents millions of downloads.

Some notable ones:
https://packagist.org/packages/twbs/bootstrap
https://packagist.org/packages/components/jquery
https://packagist.org/packages/components/jqueryui
https://packagist.org/packages/blueimp/jquery-file-upload
https://packagist.org/packages/ckeditor/ckeditor
https://packagist.org/packages/mbostock/d3
https://packagist.org/packages/tinymce/tinymce
https://packagist.org/packages/moment/moment

Tiki gets dozens of them:
http://sourceforge.net/p/tikiwiki/code/HEAD/tree/trunk/composer.json

Thanks!

arshaw:
noted. will try to get this in there when there's a spare moment

marclaporte:
Also hoping for FullCalendar Scheduler at the same time. Thanks!

arshaw:
having problems with this, as mentioned in this comment: https://github.com/fullcalendar/fullcalendar/pull/3617#issuecomment-298708728

arshaw:
still having trouble with this. see my latest comment: https://github.com/fullcalendar/fullcalendar/pull/3617#issuecomment-315505318

arshaw:
queued for release, for real this time

arshaw:
Released in v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/#more-257))

PR #3617, opened 2017-04-02
Add composer.js to let PHP developers to keep track of fullcalendar on packagist.org.

fabiomontefuscolo:
Hello!

Related to #2999 

This piece is necessary to have fullcalendar package on packagist.org. Unfortunately, there are some steps I can't do for you, like create an account in packagist.org, submit your package and wire packagist to github webhooks. But it takes less than 10 minutes.

The `name` info on composer.json should be equals to your account name and package name on packagist.org.

It would be awesome to have your library on packagist.

Thank you!!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,25 @@
+{
+    "name": "fullcalendar/fullcalendar",
+    "description": "Full-sized drag & drop event calendar",
+    "authors": [
+        {
+          "name": "Adam Shaw",
+          "email": "arshaw@arshaw.com",
+          "homepage": "http://arshaw.com/"
+        }
+    ],
+    "keywords": [
+        "calendar",
+        "event",
+        "full-sized",
+        "jquery-plugin"
+    ],
+    "type": "library",
+    "license": "MIT",
+    "minimum-stability": "stable",
+    "homepage": "https://fullcalendar.io/",
+    "require": {
+        "components/jquery": "^3.1.0",
+        "moment/moment": "^2.9.0"
+    }
+}


arshaw:

I change the jquery dependency to `2 - 3`, I hope that's okay.

I released this. Is there anything else I need to do? Does it work?

sekjal:
As of this morning, fullcalendar/fullcalendar is still NOT available on Packagist.

arshaw:
I registered it with Packagist: https://packagist.org/packages/fullcalendar/fullcalendar

However, I'm trying to get the JS/CSS files to show up in the `components` directory after `composer require fullcalendar/fullcalendar` is executed, like jquery and moment does, but I'm having a bit of trouble.

You can see my latest commits here: https://github.com/fullcalendar/fullcalendar/commits/master
To test, I'm installing the `dev-master` version in my sample composer project.

I will resume work on this soon.

Dgc2002:
Think you're almost there! I think the problem is:  
  
In the documentation for [robloach/component-installer](https://github.com/RobLoach/component-installer) there's a step called `Creating a Component`. You need to add `robloach/component-installer` to your requirements and add the configuration for the files you want installed(extra -> component ->(scripts | styles | files).

arshaw:
hmm yeah, I tried those steps and committed them to master:
https://github.com/fullcalendar/fullcalendar/blob/master/composer.json

and it still doesn't work. here's what im doing to test:

```
mkdir testproj
cd testproj
composer require robloach/component-installer
composer require fullcalendar/fullcalendar:dev-master
```

and i look in the `components` directory and it does not have the fullcalendar files. whereas the `vendor` directory does.

any ideas?

Dgc2002:
Pulling from dev-master doesn't include a 'dist' folder:  
  
image  
  
So the files defined in `extra->component->scripts|styles|files` aren't found. A bit annoying that component-installer doesn't throw an error or warning in that case :\ not even when run with the verbose flag.

arshaw:
thanks @Dgc2002 . I just tested things out with a temporary branch where I committed the built dist files and it works. When I push tagged releases, the dist files are the head commit of the tag, so all should work well when I release v3.5.0

arshaw:
Now available in Packagist as a working component after releasing v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/#more-257))

PR #3619, opened 2017-04-04
Fix toISOString for locales with non-trivial postformatting

joankaradimov:
Some locales do a post-processing step when formatting. For example Arabic does that:

```javascript
$.fullCalendar.moment.parseZone('2014-06-08T10:00:00').locale('ar').format()
// result: "٢٠١٤-٠٦-٠٨T١٠:٠٠:٠٠"
```

There is a bug in `moment-ext.js` which causes calls to `toISOString` use this post-processing:

```javascript
$.fullCalendar.moment.parseZone('2014-06-08T10:00:00').locale('ar').toISOString()
// result: "٢٠١٤-٠٦-٠٨T١٠:٠٠:٠٠"
// expected: "2014-06-08T10:00:00"
```

This PR introduces a test for that and attempts to fix the issue in a pretty ugly way.

### DIFF
--- a/src/moment-ext.js
+++ b/src/moment-ext.js
@@ -311,10 +311,10 @@ newMomentProto.format = function() {
 
 newMomentProto.toISOString = function() {
 	if (this._ambigTime) {
-		return oldMomentFormat(this, 'YYYY-MM-DD');
+		return oldMomentFormat(moment(this).locale('en'), 'YYYY-MM-DD');
 	}
 	if (this._ambigZone) {
-		return oldMomentFormat(this, 'YYYY-MM-DD[T]HH:mm:ss');
+		return oldMomentFormat(moment(this).locale('en'), 'YYYY-MM-DD[T]HH:mm:ss');
 	}
 	return oldMomentProto.toISOString.apply(this, arguments);
 };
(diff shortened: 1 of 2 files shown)


joankaradimov:
P.S.
That's weird. The tests passed locally for me. I'll take a look at it later today.

arshaw:
@joankaradimov  thanks. This has been a bug that I've heard ppl complain about, but never gotten reproduction steps for.

Something is wrong with the tests.. they are passing locally for me but on TravisCI they are failing. will look into it.

arshaw:
I've queued this up for release. I did a small optimization, and also applied the behavior to a blank `.format()` call, which should also return an ISO8601 string in my opinion, without any sort of locale.

FWIW, moment's `.toISOString()` does normalize to english, whereas `.format()` does not. But the only reason toISOString does not is because of this [flukey line which relies on an optimization](https://github.com/moment/moment/blob/2.18.1/src/lib/moment/format.js#L22).

arshaw:
im summary, after this upcoming release, any moment objects that fullcalendar creates, regardless of what language the current calendar is set to, will produce ENGLISH ISO8601 dates for both `.toISOString()` and `.format()` (called with no args).

arshaw:

Thanks!

PR #3621, opened 2017-04-05
rendering hook for columnHeader #3078, #3331, #3438

Rouche:
Old pull request closed #3447


arshaw:
it looks like you rebased all my commits from 3.3.1 . they are showing up in the diff

Rouche:
yeah, i was like 5 minutes late when i forked lol.

arshaw:
After thinking about it for a bit, here's the best API in my opinion:

```
columnHead: true/false
columnHeadFormat: date format string
columnHeadText: function
columnHeadHtml: function
```

This covers issues #3438 and #3078.

It would assume we deprecate `columnFormat`, and if the user specifies that, it would be as if they specified `columnHeadFormat` instead.

If you'd like to give that a shot, go for it. If you have any questions before beginning, please let me know.

I'll close this PR because it needs to be rebased.

Rouche:
sory @arshaw , ive never had time to implement your idea. Too much stuff hapenning right and left.

caseyjhol:
@arshaw @Rouche I might take a shot at this.

arshaw:
sounds good @caseyjhol

Issue #3615, opened 2017-03-31
Can't click on some dates. Fullcalendar not working because of overflow-x and height on html element


xxRockOnxx:
I am using the npm version of fullcalendar and I tried to recreate this in the JS Bin but I can't reproduce it there. I copied the settings from JS Bin but it just doesn't work on my build. (Read more on when it only works)

Here's what I'm trying in my build: http://output.jsbin.com/boxuqenupi/

I expect that clicking on all of the date to do `console.log()` but on mine, it doesn't log. Not only the select does not get fired but it doesn't even highlight the date at all.

I thought at first there was some conflict on other packages so I reduced my code to very basic as possible but still used the options from JS Bin:

```
import $ from 'jquery'
import 'fullcalendar/dist/fullcalendar.min.css'
import 'fullcalendar/dist/fullcalendar.min.js'

$('#calendar').fullCalendar({
    header: {
      left: 'prev,next today',
      center: 'title',
      right: 'month,agendaWeek,agendaDay,listWeek'
    },
    selectable: true,
    select: function () {
       console.log('clicked')
    },
    defaultDate: '2017-03-31',
    editable: true,
    eventLimit: true, // allow "more" link when too many events
    events: [
      {
        title: 'All Day Event',
        start: '2017-03-01'
      },
      {
        title: 'Long Event',
        start: '2017-03-07',
        end: '2017-03-10'
      },
      {
        id: 999,
        title: 'Repeating Event',
        start: '2017-03-09T16:00:00'
      },
      {
        id: 999,
        title: 'Repeating Event',
        start: '2017-03-16T16:00:00'
      },
      {
        title: 'Conference',
        start: '2017-03-30',
        end: '2017-04-01'
      },
      {
        title: 'Meeting',
        start: '2017-03-31T10:30:00',
        end: '2017-03-31T12:30:00'
      },
      {
        title: 'Lunch',
        start: '2017-03-31T12:00:00'
      },
      {
        title: 'Meeting',
        start: '2017-03-31T14:30:00'
      },
      {
        title: 'Happy Hour',
        start: '2017-03-31T17:30:00'
      },
      {
        title: 'Dinner',
        start: '2017-03-31T20:00:00'
      },
      {
        title: 'Birthday Party',
        start: '2017-04-01T07:00:00'
      },
      {
        title: 'Click for Google',
        url: 'http://google.com/',
        start: '2017-03-28'
      }
    ]
  });

```

[...]

PR #3627, opened 2017-04-09
Added hack fix for jquery-ui selectable bug

xxRockOnxx:
Bug fix for this issue I created: https://github.com/fullcalendar/fullcalendar/issues/3615

Bin: http://jsbin.com/kufidajona/edit?output

Actual issue is, some dates are not clickable/selectable because of a jquery-ui bug.

Sometimes just the `overflow-x` alone can result in this bug and sometimes (like in the Bin) needs the `height: 100%` on `html` and `body`

Others will end up doing it this way anyway, why not add it in the core? I suffered for 1 week because of this lol.

Sounds weird but hey... it's a hack. Can't even explain why.

### DIFF
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -14,6 +14,10 @@
 	text-align: right;
 }
 
+body { /* jquery-ui selectable bug HACK fix */
+	overflow-x: visible !important!;	
+}
+
 body .fc { /* extra precedence to overcome jqui */
 	font-size: 1em;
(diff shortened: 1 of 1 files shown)


arshaw:
i don't want fullcalendar to be responsible for providing a hack to reverse a bug from another library.  i appreciate the pragmatism, but hopefully the jquery ui guys will fix that.

PR #3630, opened 2017-04-11
Use a dynamically set duration object if available

bjuffe:
Right now setting a duration dynamically is ignored. Example below.
`$('#calendar').fullCalendar('option', 'duration', { weeks: 10 })` 

This change will use the dynamic override in the first place.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -140,8 +140,8 @@ var Calendar = FC.Calendar = Class.extend({
 		}
 
 		// fall back to top-level `duration` option
-		durationInput = durationInput ||
-			this.dynamicOverrides.duration ||
+		durationInput = this.dynamicOverrides.duration ||
+			durationInput ||
 			this.overrides.duration;
 
 		if (durationInput) {


arshaw:
thanks!

arshaw:
@bjuffe I'm further considering this and not sure I want a dynamic setter for `duration` to affect the duration all views, which is what your change would enact. I only want a dynamic setter to affect a generic view (a view without an explicitly defined initial duration) rather than affecting standard views, like month view. Same for custom views.

So, your code is probably like this:

```js
$('#calendar').fullCalendar({
  views: {
    flexibleView: {
      type: 'agenda',
      duration: { days: 3 } // won't change
    }
  }
});
$('#calendar').fullCalendar('option', 'duration', { weeks: 1 });
```

I'm proposing you change it to something like this:

```js
$('#calendar').fullCalendar({
  duration: { days: 3 }, // the default duration for flexible views
  views: {
    flexibleView: {
      type: 'agenda'
    }
  }
});
$('#calendar').fullCalendar('option', 'duration', { weeks: 1 });
```

Does that make sense?

bjuffe:
@arshaw, you are absolutely right, it makes total sense. Thanks for the help! Should this be added to the documentation for dynamic setters? I'll be happy to help.

arshaw:
good idea, i made a ticket to modify the documentation: https://github.com/fullcalendar/fullcalendar/issues/3659

PR #3634, opened 2017-04-13
Create yi.js

zvizvi:
Add Yiddish language


arshaw:
thanks for the submission. moment.js also needs to support Yiddish for this to work, and it doesn't look like they do...
https://github.com/moment/moment/tree/develop/locale

PR #3649, opened 2017-04-25
Set isComputed = true when scrollTime option is set

compleatguru:
when scrollTime is set, I noticed that the scroll bar will move to the set time based on 
```
releaseScroll: 

...
if(scroll.isComputed)
```

However, the scroll object depends on this.capturedScroll(), which depends on this.queryScroll()

this.queryScroll() will never return an object with isComputed property.

Thus, I am making this fix so that the scroll bar will not be reset to left:0, top:0.


arshaw:
@compleatguru , the scrolling system has been overhauled in v3.4.0 and this bug is likely fixed. could you try out the newest version please?

PR #3656, opened 2017-04-27
Serbian (cyrlic) translation [skip ci]

nemanja947:
Translation of interface in serbian cyrilic.

### DIFF
--- a/locale/sr-cyrl.js
+++ b/locale/sr-cyrl.js
@@ -1,0 +1,14 @@
+
+$.fullCalendar.locale("sr", {
+	buttonText: {
+		month: "Mеsеc",
+		week: "Nеdеlja",
+		day: "Dan",
+		list: "Planеr"
+	},
+	allDayText: "Cеo dan",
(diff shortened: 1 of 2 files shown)


arshaw:
Thanks.

The remaining work is on my end because the jQuery UI translation file is named "datepicker-sr-SR.js" ([here](https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-sr-SR.js)) and fullcalendar will confused while compiling.

joankaradimov:
> translation file is named "datepicker-sr-SR.js"

Just a note - that does not appear correct on jQuery UI's side. While the language code is `sr`, the country code is `RS` as per [ISO 3166](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). Also `.rs` is Serbia's top level domain.

arshaw:
Released in v3.5.0 ([blog post](https://fullcalendar.io/blog/2017/08/bootstrap-support-and-future-developments/#more-257))

PR #3397, opened 2016-10-28
fullcalendar-Jalaali 3.0.1

Natico:
change and update many files for support fullcalendar -jalaali 3.0.1

PR #3680, opened 2017-05-17
Add Jalali support

sijav:
Hi, this is a suggestion for adding jalali support for full-calendar.
I've taken ideas and codes from [#3397](https://github.com/fullcalendar/fullcalendar/pull/3397) and changed it in a way so it can be a clean pr and works with 3.4 although this might needs some extra works, I'm not aware of any bugs with this so if you encounter any let me know.

### DIFF
--- a/src/date-formatting.js
+++ b/src/date-formatting.js
@@ -199,11 +199,16 @@ function renderParsedFormat(parsedFormat, date1, date2, separator, isRTL, isJala
 	}
 
 	if (middleStr1 || middleStr2) {
-		if (isRTL) {
-			middleStr = middleStr2 + separator + middleStr1;
+		if (middleStr2 == middleStr1) {
+				middleStr = middleStr1;
 		}
 		else {
-			middleStr = middleStr1 + separator + middleStr2;
+			if (isRTL) {
+				middleStr = middleStr2 + separator + middleStr1;
+			}
+			else {
+				middleStr = middleStr1 + separator + middleStr2;
+			}
 		}
 	}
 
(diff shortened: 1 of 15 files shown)


src/date-formatting.js
```diff
@@ -186,9 +226,9 @@ var parsedFormatStrCache = {};
 /*
 Returns a parsed format string, leveraging a cache.
 */
-function getParsedFormatString(formatStr) {
+function getParsedFormatString(formatStr, isJalaali) {
 	return parsedFormatStrCache[formatStr] ||
```
sijav: there's a bug here, the buildSameUnits function does not work as expected because jalali calendar is based on dd mm yy while georgian calendar is based on mm dd yy ... there must be fix for it, I'm working on it

sijav: parsedFormatStrCache is only for format so it can conflict with default format ... I'll try to create separate cache for this

sijav:
Well come to think of it it would be better idea to create a calcType variable which state that which calclType should be choose and then replace the isJalaali with calcType variable and check which calcType it is to use them accordingly ...

src/date-formatting.js
sijav: I've made some progress here but still encounter some minor issues, So I'm gonna still hold on a little bit longer to see if there's any better idea to what to do here ...

sijav:
Is this project still being maintained? Can you review this pr please?

arshaw:
thanks for this PR @sijav , though I won't merge this PR into the main branch for the reasons I list here: https://github.com/fullcalendar/fullcalendar/issues/1813#issuecomment-281620201

basically, this is blocked on momentjs adding support for multiple calendar systems. i'd rather do it that way than rely on moment-jalaali, which is essentially a monkeypatched version of moment.

arshaw:
hi @sijav. thank you for making this PR a while ago. aside from the codebase no longer being based on moment, i've come up with a new technique that I aim to use when implementing Jalaali support. See my [latest comment here](https://github.com/fullcalendar/fullcalendar/issues/1813#issuecomment-652600319). So, I'm closing this PR, hope you understand

PR #3685, opened 2017-05-19
Use the week with the Monday when ISO week calculation

laoneo:
When the start date of the week is set to Sunday and the week number calculation is set to ISO, then the weeknumber is shown of the week from the Sunday.
Does it make sense to show the week number of the first Monday in the row?

### DIFF
--- a/src/basic/BasicView.js
+++ b/src/basic/BasicView.js
@@ -383,6 +383,11 @@ var basicDayGridMethods = {
 		var view = this.view;
 		var weekStart = this.getCellDate(row, 0);
 
+		// If week number calc is ISO, then we need to use the week of Monday
+		if (weekStart._locale._fullCalendar_weekCalc === 'ISO') {
+			weekStart = weekStart.day(1);
+		}
+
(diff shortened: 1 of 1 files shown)


arshaw:
I'd hate to put another week number edge case here.

Is this a scenario that people want? Wouldn't the vast majority of people who want to display ISO week numbers also have their weeks begin with Monday?

laoneo:
A user of my Joomla extension had it the way that he wanted to show the week numbers in ISO format but the week was starting on Sunday. According to wikipedia you are right, the ISO week starts with Monday https://en.wikipedia.org/wiki/ISO_week_date.

jjnxpct:
Hi! We would also like to be able to start the week on monday but have the correct weeknumber in front of the week in month view. Would it be possible to use merge the pull request by laoneo? We use his calendar extension in Joomla. Or does the pull request have a conflict that need to be resolved first?

arshaw:
this PR has been superseded by https://github.com/fullcalendar/fullcalendar/pull/4731

PR #3695, opened 2017-06-01
Add translations for list views

kmukku:
All list views has same title, so I add some translations to separate the views.

### DIFF
--- a/locale/fi.js
+++ b/locale/fi.js
@@ -4,7 +4,11 @@ $.fullCalendar.locale("fi", {
 		month: "Kuukausi",
 		week: "Viikko",
 		day: "Päivä",
-		list: "Tapahtumat"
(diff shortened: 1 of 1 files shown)


arshaw:
@kmukku thanks for the change. I'm not sure I want every locale to need to specify each one of these. I'm hoping that when list view endless scroll is released (might be a while off), people will begin to use one single list view for all times spans and this becomes a moot point.

I'll pass on accepting this change for now, but you can always do this yourself on the main config like this:

```js
views: {
	listYear: { buttonText: "Vuoden tapahtumat" },
	listMonth: { buttonText: "Kuukauden tapahtumat" },
	listWeek: { buttonText: "Viikon tapahtumat" },
	listDay: { buttonText: "Päivän tapahtumat" }
}
```

PR #3706, opened 2017-06-03
Create Sebian Latin locale

ergonomicus:
Created sr-Latn.js for Serbian Latin version.

### DIFF
--- a/locale/sr-Latn.js
+++ b/locale/sr-Latn.js
@@ -1,0 +1,14 @@
+
+$.fullCalendar.locale("sr-Latn", {
+  buttonText: {
+    month: "Mesec",
+    week: "Nedelja",
+    day: "Dan",
+    list: "Planer"
+  },
+  allDayText: "Ceo dan",
+  eventLimitText: function(n) {
+    return "+ još " + n;
+  },
+  noEventsMessage: "Nema događaja za prikaz"
+});


arshaw:
I think most libraries simply use "sr" as the serbian latin abbreviation, no? could you rename? that's the only way the build system will know to find the corresponding jquery ui and moment files:

- https://github.com/moment/moment/blob/develop/locale/sr.js
- https://github.com/jquery/jquery-ui/blob/master/ui/i18n/datepicker-sr.js

Could you build the translation and ensure it works?

arshaw:
apologies. i just realized this existing PR is already handling this task:
https://github.com/fullcalendar/fullcalendar/pull/3656

@nemanja947 correctly repurposed "sr.js" be the latin and "sr-cyrl.js" be the cyrilic. But it's blocked on fullcalendar's i18n build system.

ergonomicus:
Well, there is no consistency regarding this. Serbian is [digraphic](https://en.wikipedia.org/wiki/Digraphia) language and both Latin and Cyrillic are equally used.

If you look at **moment** implementation they have **sr.js** referring to Latin script, and **sr-cyrl.js** referring to Cyrillic script.

Date picker uses 'default' Cyrillic in **datepicker-sr.js**, but has odd **datepicker-sr-SR.js** for Latin.

I guess that this will be resolved someday, until then - we're on our own :-)

Issue #3732, opened 2017-06-24
Problem with prev[Year], next[Year] and dynamic validRange


BenVercammen:
Code: http://jsfiddle.net/wytj8z5u/
Issue:
- Set the validRange dynamically (using a callback function)
- Go back in time using the prevYear button
- The validRange prevents from going back more than 3 months in the past
- Issue 1: The prevYear buttons is not disabled
- Issue 2 (more important): The "next" button is disabled

For now, users will have to click nextYear or today button to enable the next button again.

Probably related to https://github.com/fullcalendar/fullcalendar/issues/3651

arshaw:
recreated, thanks

bestimpressions-dev:
I also found similar issue with prev_month when range is dynamically changed. 

Prev month start to be enabled when you dynamically set the same range as on init. check this code:  http://jsfiddle.net/eckma2hm/1/

PR #3774, opened 2017-08-01
Added enabling/disabling prevYear and nextYear buttons

paw3lx:
Related to #3732 

I added disabling/enabling previous and next year buttons, depending on validRange. I think they should work in the same way as prev/next buttons.

The other problem with disabling next button is already solved.

### DIFF
--- a/src/Calendar.toolbar.js
+++ b/src/Calendar.toolbar.js
@@ -87,6 +87,20 @@ Calendar.mixin({
 				'disableButton',
 			'next'
 		);
+
+		this.toolbarsManager.proxyCall(
+			prevInfo.isValid ?
+				'enableButton' :
+				'disableButton',
+			'prevYear'
+		);
+
+		this.toolbarsManager.proxyCall(
+			nextInfo.isValid ?
+				'enableButton' :
+				'disableButton',
+			'nextYear'
+		);
 	},
 
 


src/Calendar.toolbar.js
```diff
@@ -87,6 +87,20 @@ Calendar.mixin({
 				'disableButton',
 			'next'
 		);
+
+		this.toolbarsManager.proxyCall(
+			prevInfo.isValid ?
```
arshaw: the `prevInfo` and `nextInfo` objects you are leveraging here are information for the prev and next click (for example, one month ahead or back), not necessarily one year ahead or one year back. probably fine in 99% of usecases, but would prefer if you generated and tested a date profile for one year forward/back explicitly.

paw3lx: Ok, I understand. But in my opinion the date profile for one year forward will be the same as prev date profile builded from buildPrevDateProfile method, right?
Or maybe I don't understand when prev and next year buttons should be enabled/disabled. I think that prevYear button should be enabled when you can go back at least one day, right? Similarly with nextYear.

arshaw: @paw3lx that would be a convenient shortcut, but no, sometimes the user could go back a day but could NOT go back a year due to the [validRange](https://fullcalendar.io/docs/current_date/validRange/) limiting that large of a jump in navigation

arshaw:
I'm closing this PR because it has gotten way out of sync with the main codebase (my fault), but I created this new issue to track demand for it: https://github.com/fullcalendar/fullcalendar/issues/5575

PR #3788, opened 2017-08-11
Add agendaEventMinHeight option to prevent short events overlapping

Stafie:
Issue: https://github.com/fullcalendar/fullcalendar/issues/961

Fiddle illustrating the problem: http://jsfiddle.net/jc1zx2q6/5/
Fiddle illustrating the solution: https://plnkr.co/edit/t5eAi1raDw4ZexwCrWy6?p=preview

### DIFF
--- a/src/agenda/TimeGrid.js
+++ b/src/agenda/TimeGrid.js
@@ -471,13 +471,14 @@ var TimeGrid = FC.TimeGrid = InteractiveDateComponent.extend(StandardInteraction
 	computeSegVerticals: function(segs) {
 		var i, seg;
 		var dayDate;
+		var agendaEventMinHeight = this.view.opt('agendaEventMinHeight');
 
 		for (i = 0; i < segs.length; i++) {
 			seg = segs[i];
 			dayDate = this.dayDates[seg.dayIndex];
 
 			seg.top = this.computeDateTop(seg.startMs, dayDate);
-			seg.bottom = this.computeDateTop(seg.endMs, dayDate);
+			seg.bottom = Math.max(seg.top + agendaEventMinHeight, this.computeDateTop(seg.endMs, dayDate));
 		}
 	},
 
--- a/src/View.js
+++ b/src/View.js
@@ -31,6 +31,7 @@ var View = FC.View = InteractiveDateComponent.extend({
 	nowIndicatorTimeoutID: null, // for refresh timing of now indicator
 	nowIndicatorIntervalID: null, // "
 
+	agendaEventMinHeight: 0, // allows setting a min-height to the event segment to prevent short events overlapping each other
 
 	constructor: function(calendar, viewSpec) {
 		this.calendar = calendar;
--- a/tests/event-render/minSegmentHeight.js
+++ b/tests/event-render/minSegmentHeight.js
@@ -1,0 +1,50 @@
+describe('short event rendering with agendaEventMinHeight', function() {
+	pushOptions({
+		defaultView: 'agendaWeek',
+		defaultDate: '2017-08-10',
+		agendaEventMinHeight: 25
+	});
+
+	describe('we we have an isolated short event', function() {
+		pushOptions({
+			events: [
+				{ start: '2017-08-10T10:30:00', end: '2017-08-10T10:31:00' }
+			]
+		});
+
+		it('renders the event having full width and the agendaEventMinHeight height value', function() {
+			initCalendar();
+			var el = EventRenderUtils.getSingleEl();
+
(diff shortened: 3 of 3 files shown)


arshaw:
This is a good stopgap solution, using a JS setting, but I would prefer if the CSS was able to specify the min-height. In other words, I'd prefer if the DOM was consulted for the min-height.

I might need to merge this tho, to get it out there faster. Could you rename the setting to `agendaEventMinHeight` ?

src/agenda/TimeGrid.js
```diff
@@ -477,7 +477,12 @@ var TimeGrid = FC.TimeGrid = InteractiveDateComponent.extend(StandardInteraction
 			dayDate = this.dayDates[seg.dayIndex];
 
 			seg.top = this.computeDateTop(seg.startMs, dayDate);
-			seg.bottom = this.computeDateTop(seg.endMs, dayDate);
+			if(this.computeDateTop(seg.endMs, dayDate) - seg.top < this.view.opt('minSegmentHeight')) {
```
arshaw: could you cache the value of the `opt()` call in a variable at the top of the function?

src/agenda/TimeGridEventRenderer.js
```diff
@@ -273,6 +273,7 @@ var TimeGridEventRenderer = EventRenderer.extend({
 		props.zIndex = seg.level + 1; // convert from 0-base to 1-based
 		props.left = left * 100 + '%';
 		props.right = right * 100 + '%';
+		props.minWidth = 'calc(' + (forwardCoord - backwardCoord) * 100 + '%' + ' - 2px)';
```
arshaw: what is this doing?

Stafie: I remember having some issues with events that had short width, but I couldn't replicate it now so I removed this.

src/agenda/TimeGrid.js
Stafie: done.

Stafie:
I apologize for the delay, I made the adjustments you asked for.

src/agenda/agenda.css
```diff
@@ -217,6 +217,11 @@ be a descendant of the grid when it is being dragged.
 
 /* short mode, where time and title are on the same line */
 
+.fc-time-grid-event.fc-short:hover {
```
arshaw: why was this added? if it's not related to the main feature in this PR, could you please remove it?

src/agenda/TimeGrid.js
```diff
@@ -471,13 +471,19 @@ var TimeGrid = FC.TimeGrid = InteractiveDateComponent.extend(StandardInteraction
...
 		var dayDate;
+		var agendaEventMinHeight = this.view.opt('agendaEventMinHeight');
 
 		for (i = 0; i < segs.length; i++) {
 			seg = segs[i];
 			dayDate = this.dayDates[seg.dayIndex];
 
 			seg.top = this.computeDateTop(seg.startMs, dayDate);
-			seg.bottom = this.computeDateTop(seg.endMs, dayDate);
+			if(this.computeDateTop(seg.endMs, dayDate) - seg.top < this.view.opt('agendaEventMinHeight')) {
```
arshaw: please us the `agendaEventMinHeight` you created above. also, you are calling `computeDateTop` twice. could you also please put that in a variable.

Also, please do this without an `if` statement. That should be possible with `Math.max`

Stafie: Made the changes here.
Since we got rid of the `if` statement, now we use the `computeDateTop` once, so didn't have to put it in a variable.

src/agenda/agenda.css
Stafie: This allows to view the event data when they are crammed together on the same line. GIF:
https://www.dropbox.com/s/b512d3xweimrwz6/event%20hovering.gif?dl=0

Let me know if you want to get rid of it, and I'll do a fixup.

arshaw: yes, could you get rid of it and put it into a different PR. it's a distinct enough feature, and up for debate as to what the exact API should be, and if it should be the new default.

Stafie: Got rid of it.

arshaw:
thanks

arshaw:
queued for release. i did a fix or two

arshaw:
Released in [v3.6.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.6.0)

docs: https://fullcalendar.io/docs/agenda/agendaEventMinHeight/

Issue #1580, opened 2015-08-19
iCal (ics) support


arshaw:
Originally reported on Google Code with ID 1313

```
Would be great if we could import ics files or if there was some documentation on how
to do so.
```

Reported by `andrewmcmurry` on 2012-02-22 23:57:06

Imported with **19** stars.

2733:
Hi, i wrote this https://github.com/2733/icalendar2fullcalendar.
Dunno if it could be useful to integrate it in FullCalendar.

kingcheez:
Thanks. Does it work for recurring events?
 
On Thu, Mar 10, 2016, at 10:36 PM, 2733 wrote:
> Hi, i wrote this https://github.com/2733/icalendar2fullcalendar.
>
Dunno if it could be useful to integrate it in FullCalendar.
> —
> Reply to this email directly or view it on GitHub[1].
>
 

Links:

  1. https://github.com/fullcalendar/fullcalendar/issues/1580#issuecomment-195164076

2733:
No, because fullcalendar doesnt support recurring events, does it?
I can workaround that by generating one-time events from a recurring event.
Any advice?

kingcheez:
Not exactly, but ical.js exposes an API for evaluating recurrences over a period:
 
https://mozilla-comm.github.io/ical.js/api/ICAL.RecurExpansion.html
 
This could be used in combination with the start/end parameters from the
'events' callback in fullcalendar and made to work.

2733:
Now recurring events work too.

kingcheez:
This seems overly complex (two interface functions when I think only one is needed) but nice to see that it works. I might have an opportunity to test it later. Thanks for doing the work!

On March 16, 2016 12:01:18 PM EDT, 2733 <notifications@github.com> wrote:
>Now recurring events work too.
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/fullcalendar/fullcalendar/issues/1580#issuecomment-197397629

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

2733:
Please report this and other issues on my repo, not to go off-topic here.

DoobleD:
Hi, any news on this new feature ? Not implemented yet ?

pokoli:
@ghost do you plan to create a PR on this repo to integrate it as base feature?

larrybolt:
[...]

PR #3805, opened 2017-08-25
Implemented iCalendar/ICS Event Source

haukepribnow:
This pull request includes an implementation of an Event Source that supports the iCalendar/ICS standard. It is currently based on [a fork of `ical-expander`](https://github.com/haukepribnow/ical-expander) that I created to make `ical-expander` a Universal JavaScript Module. My fork and [the original `ical-expander`](https://github.com/mifi/ical-expander) are based on [`ical.js`](https://github.com/mozilla-comm/ical.js).

[I hope that my forked branch will be pulled into the original `ical-expander` project at some point.](https://github.com/mifi/ical-expander/pull/2) Once this happens, I would also change the `ical-expander` dependency from the GitHub source into the "normal" package source.


arshaw:
hi @haukepribnow this is pretty cool. Both `EventSource` and `EventSourceParser` are exposed

https://github.com/fullcalendar/fullcalendar/blob/v3.5/src/models/event-source/EventSource.js#L155
https://github.com/fullcalendar/fullcalendar/blob/v3.5/src/models/event-source/EventSourceParser.js#L29

and you can get ahold of them like this:

```js
var EventSource = $.fullCalendar.EventSource;
```

so you could easily write this as a plugin for FullCalendar and not have it be part of the core repo, which would be preferable, at least at the start.

Does that make sense?

haukepribnow:
Thank you for your feedback. Turning iCalendar integration in fullcalendar into a plugin for fullcalendar makes sense to me.

On a different note: The author of `ical-expander` stated in the meantime that [he does not want to turn his project into a Universal JavaScript Module](https://github.com/mifi/ical-expander/pull/2#issuecomment-325956515). Instead, he suggests to use Browserify or webpack.

Now I wonder what the best solution would be for a fullcalender iCalendar plugin as suggested by you. Would you recommend to use Browserify or webpack for the plugin? Are you maybe considering to move your own fullcalendar project towards one of the two?

After all, I wonder a bit if it makes sense to introduce Browserify or webpack in my fullcalendar plugin, while fullcalendar remains a Universal JavaScript Module without using any of the two. Does that really make sense? 

To be honest, this confusion might come from the fact that I have been more of a "user" of modules such as yours so far. So I'm not really into the ecosystem of and npm modules and packages yet. (Coming more from a C# background.) So any hints on how to proceed now would be greatly appreciated. :-)

arshaw:
@haukepribnow, fullcalendar is implemented as a UMD module:
https://github.com/fullcalendar/fullcalendar/blob/v3.5.0/dist/fullcalendar.js#L7-L17

UMD modules are meant to work with Browserify and Webpack.

But since your main dependency (ical-expander) is not written as a UMD, I suggest implementing your plugin as an ES6 modules (which works with Browserify and Webpack). It can consume fullcalendar all the same.

The exact steps are rather complicated :(
https://www.smashingmagazine.com/2016/02/writing-reusable-components-es6/

arshaw:
I'm closing this PR in favor of https://github.com/fullcalendar/fullcalendar/pull/5969, which has been released in v5.5.0. Please see my comment here: https://github.com/fullcalendar/fullcalendar/issues/1580#issuecomment-748535186

PR #3807, opened 2017-08-29
Fix for the gitignore, and the view

ParanormalCoder:
This fix provides a possibility to stop the reset scroll for the agenda.

The fixes are provided as follows:

1. .gitignore is augmented
2. AgendaView is facilitated with the globalForcedScroll condition
3. Calendar render mixin is changed to mention the globalForcedScroll for the isForced state of the view
4. globalForcedScroll can be used for preventing the Agenda scroll reset

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,13 @@ dist
 # npm
 node_modules
 npm-debug.log
+package-lock.json
+
+# yarn
+yarn.lock
+
+# JetBrains
+.idea/
+
+# osx
+.DS_Store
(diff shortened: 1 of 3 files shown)


arshaw:
Hi @ParanormalCoder , what's the "reset scroll" for agenda view? Could you elaborate more? Is there an issue tracker ticket for this?

ParanormalCoder:
Hi @arshaw,

I will create an issue tracker for this required feature. In brief, the calendar needs to keep the scroll position while user changes/switches the day. I will describe it more inside the issue tracker.

PR #3815, opened 2017-09-01
V3.5: Add support for reversed year/month lists (by date)

devBluestar:
**Support for a new configuration flag:**

$('#calendar').fullCalendar({
     eventListReverse: true  { "false" by default }
}

[fullcalendar.zip](https://github.com/fullcalendar/fullcalendar/files/1270579/fullcalendar.zip)


espen:
Looks like you pushed the incorrect commits?

danielfaust:
@devBluestar When I go to "Files changed (21)" and do a search for `eventListReverse` not a single match shows up. The zipfile seem to contain your changes regarding `eventListReverse`, I will see if it works for me.

Proffitt78:
I tried using the file in that zip folder... it worked for the calendar that was in list view. But it broke the other calendar I have on the page that is using the default view.

avital-rubin:
thank you for a great solution.
i didnt take the all .js file, just copied the additional part and it works perfectly

PR #3833, opened 2017-09-12
fix #3341 (unselect callback return order)

### LINKED ISSUE #3341: Unselect Callback Return Order
Per documentation: https://fullcalendar.io/docs/selection/unselect_callback/
Expected: function(view, jsEvent)
Actual: function(jsEvent, view)

Reproduced: http://jsbin.com/yequqolufu/1/edit?html,js,console,output

Easiest fix is updating the documentation to read function(jsEvent, view)? since convention seems to return jsEvent before view on other callbacks.

caseyjhol:
Takes care of #3341.

### DIFF
--- a/selection/unselect_callback.txt
+++ b/selection/unselect_callback.txt
@@ -5,7 +5,7 @@ unselect
 A callback that will fire when the current selection is cleared.
 
 <div class='spec' markdown='1'>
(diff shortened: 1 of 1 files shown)


arshaw:
live: https://fullcalendar.io/docs/selection/unselect_callback/

PR #3834, opened 2017-09-13
Sort event on duration is configurable by isDurationSort boolean. By default it is true

aqil-tariq-confiz:
Sort event on duration is configurable by isDurationSort boolean. By …default it is true.
For further reading read this
https://stackoverflow.com/questions/46052484/order-events-in-fullcalendar-by-start-date-and-end-date
https://stackoverflow.com/questions/46150370/eventorder-property-of-fullcalendar-io-is-not-working-correctly

### DIFF
--- a/src/View.js
+++ b/src/View.js
@@ -20,6 +20,7 @@ var View = FC.View = InteractiveDateComponent.extend({
 	selectedEventInstance: null,
 
 	eventOrderSpecs: null, // criteria for ordering events when they have same date/time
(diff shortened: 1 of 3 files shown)


arshaw:
thanks @aqil-tariq-confiz! you or someone else would need to write some automated tests for this before merging. https://github.com/fullcalendar/fullcalendar/wiki/Automated-Tests

arshaw:
v4 changed the meaning of `eventOrder` to allow this
https://fullcalendar.io/docs/eventOrder

PR #3838, opened 2017-09-15
Check for console.log/warn.apply

jacobmischka:
Thanks to IE it seems it's possible for `console.log`/`console.warn` to exist without the `apply` method.

image

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -661,7 +661,7 @@ function isTimeString(str) {
 FC.log = function() {
 	var console = window.console;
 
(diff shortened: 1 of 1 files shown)


jacobmischka:
Hm alright apparently that part hasn't been touched for 2 years and IE9 definitely used to work so I'm probably running into a different problem.

jacobmischka:
Alright yeah, not sure what IE was doing but just restarting the VM worked. Sorry!

PR #3847, opened 2017-09-25
Creating sq.js for the Albanian language

alensaqe:
Under ISO standards Albanian two letter code is 'sq'

### DIFF
--- a/locale/sq.js
+++ b/locale/sq.js
@@ -1,0 +1,13 @@
+$.fullCalendar.locale("sq", {
+	buttonText: {
+		month: "Muaj",
+		week: "Javë",
+		day: "Ditë",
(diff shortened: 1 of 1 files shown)


arshaw:
thanks

alensaqe:
One question. Is locale-all.js auto generated? 

As it seems way more complicated from what i just translated i know you use moment.js localization for that matter.

For now i added more data into our local sq.js file, so it properly displays some more translation sets like months/weekdays etc.

arshaw:
Released in [v3.6.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.6.0)

alensaqe:
That's great. Have a nice day!

PR #3848, opened 2017-09-26
Set 'this' context properly in FuncEventSource

declspec:
The current code sets the `this` context when invoking the user-defined event source function. However it uses `this.calendar`, which is undefined in the `Promise` callback. It should be using `_this` to access the correct calendar object.

The default implementation is a fairly innocuous bug and won't cause any issues unless the caller tries to use `this` in their event source function. However, if you're in `use strict` mode (i.e. you've included `fullcalendar` as part of a build-chain) the browser will throw errors and stop execution.

https://jsbin.com/sepeparuse/edit?html,js,console,output

### DIFF
--- a/src/models/event-source/FuncEventSource.js
+++ b/src/models/event-source/FuncEventSource.js
@@ -9,7 +9,7 @@ var FuncEventSource = EventSource.extend({
 
 		return Promise.construct(function(onResolve) {
 			_this.func.call(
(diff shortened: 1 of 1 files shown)


arshaw:
thanks

arshaw:
released in 3.6.0

declspec:
Awesome, cheers!

PR #3862, opened 2017-10-09
Update af.js

ettiennelouw:
Updated no event message to be more correct wording


arshaw:
released in https://github.com/fullcalendar/fullcalendar/releases/tag/v3.8.0
thanks!

PR #3875, opened 2017-10-16
Guard against missing segment elements

dahlbyk:
Since upgrading to 3.6, I'm hitting `Cannot read property 'css' of undefined` errors for some segments, which I have confirmed are indeed missing `el`. The patch is trivial, but I can pin down a repro if necessary.


arshaw:
@dahlbyk yes can you please post a repro. I want to know more about the underlying bug

Issue #387, opened 2015-08-19
Recurring Events


arshaw:
Originally reported on Google Code with ID 116

```
Hi,

 It will be nice that we can add recurring events. 
 I was trying to use this plugin to display birthdates. 
 To do so i had to generate 101 events (50 year back and 50 year forward) to 
achieve the kind of display i was looking for (i expect that no one will do 
50 clicks in either directions :) ), which is not very nice.

Thanks in advance

```

Reported by `rui.estevao` on 2009-10-06 15:35:42

Imported with **46** stars.

sulian:
@arshaw , slicedtoad from stackoverflow give a good answer about this problem : http://stackoverflow.com/a/29393128

johndave8116:
Hi All,

My Requirement: I have added an event as:
Start Date: 1 Jan 2016
End Date: 31 Dec 2016
Start Time: 10 AM
End Time: 11 AM
Repeat every: 2 Weeks
Repeat Days: Monday, Friday, Sunday

I have found http://fullcalendar.io/docs/display/businessHours/
businesshours for recursive for every week. But I need biweekly.

How can this be achieved?

Please help me urgently.

Thank you.

lellisga:
The calendar is duplicating events ( see the attachment ) If I click the calendar, create a new recurrent event and then click on 'Next' or 'Prev' ( It displays normally until I click on Next of Prev).

![screen shot 2016-10-10 at 4 04 59 pm](https://cloud.githubusercontent.com/assets/210771/19251122/583ed546-8f03-11e6-8f8d-4e84779c9659.png)

This is the event that I'm creating:

```js
var event = {
         start: '10:00',
         end:   '14:00',
         "dow": [
           "1",
           "3",
           "4"
         ],
         "title": "OUT",
         "color": "Grey"
       };
```

Here's the complete code:

```html
<html>
  <head>
    <link rel="stylesheet" href="bower_components/fullcalendar/dist/fullcalendar.css" />
    <script src="bower_components/jquery/dist/jquery.js"></script>
    <script src="bower_components/moment/moment.js"></script>
    <script src="bower_components/fullcalendar/dist/fullcalendar.js"></script>
  </head>

  <body>

    <div id="fullCalendar"></div>

    <script type="text/javascript">
     var calendarOptions = {
       header: {
         left:   'title',
         center: '',
         right:  'today prev,next'
       },
       defaultView:  'agendaWeek',
       editable:     true,
       selectable:   true,
[...]

PR #3879, opened 2017-10-17
Add a callback filtering instantiated events of business hours

daigot:
Hello,

I'm using businessHours setting.
And I want to add irregular (non-periodic) non business days (e.g. public holidays) and business days (e.g. extraordinary opening).
But currently FullCalendar's businessHours setting allows only dow.

This pull request adds a filter callback which enable adding/removing elements to/from instantiated events of businessHours.

Thanks.

### DIFF
--- a/src/models/BusinessHourGenerator.js
+++ b/src/models/BusinessHourGenerator.js
@@ -24,16 +24,17 @@ var BusinessHourGenerator = FC.BusinessHourGenerator = Class.extend({
 		var eventDefs = this.buildEventDefs(isAllDay);
 		var eventInstanceGroup;
 		var eventInstances;
+		var buildInstancesFunc = $.proxy(this.buildSingleEventInstances, this, isAllDay, unzonedRange);
 
 		if (eventDefs.length) {
 			eventInstances = eventDefsToEventInstances(eventDefs, unzonedRange);
 
 			if (this.calendar.hasPublicHandlers('businessHourEventFilter')) {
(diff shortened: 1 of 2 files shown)


arshaw:
[Background Events](https://fullcalendar.io/docs/event_rendering/Background_Events/) are your best bet for this. They can be programmatically generated unlike businesshours. you can get the same styling if you set `render: 'inverse-background'` and `className: 'fc-nonbusiness'`

When [repeating-event support](https://github.com/fullcalendar/fullcalendar/issues/387) improves, businesshours will be able to leverage all the same functionality, so will become a moot point.

daigot:
@arshaw , thank you for your reply.

I had tried background rendering with 'fc-nonbusiness' before.
But, this solution (background event) has several problems for me.

https://jsfiddle.net/daigot/5u648q4x/

1. Background events are overlaied on regular businessHours. (intersection become darker)
2. constraint: 'businessHours' has no effect on background events.
3. Every background events need two event definitions for all-day slot and the others.

Currently, are there any solutions for these problems ?

I hope businessHours improves like recurring events support.

I'm sorry for taking your time.
Thanks.

arshaw:
1) don't use a combination of `businessHours` and background events. use background events all the way.
2) put an ID within the event definitions of your background events and set `eventConstraint` to that ID. should work the same way.
3) ya, I know it's not ideal

daigot:
Thank you.
I understand.

Though it is a bit hard that creating events from dow and irregular business/non-business days, I'll try to use inverse-background events for the present.

Issue #3872, opened 2017-10-13
Now Indicator arrow not properly updated since 3.6


neroux:
Since 3.6 the _Now Indicator arrow_ does not get updated properly anymore after changing dates. It appears the way how `.fc-content-skeleton` containers are created was changed and the previous one does not get replaced anymore upon a change but the new one is simply appended.

In this context it seems the `.fc-now-indicator-arrow` element is added to all those containers now but only removed from the first (most recent?) in the DOM.

neroux:
@arshaw Re-opened as I can now actually reproduce it. Still works under 3.5.1, seems to have been introduced with 3.6.

neroux:
@arshaw From what I could debug so far, it seems the problem is a few layers above the arrow element. The `div.fc-content-skeleton` element seems to be duplicated on each period change since 3.6.

dhiemstra:
Also have this issue. FC doesn't remove the element from the DOM at all.

![image](https://user-images.githubusercontent.com/37780/31770037-6abc9100-b4d6-11e7-9ca3-590cb38f11e0.png)

arshaw:
thanks, I'll take a look at this for the next bugfix release

arshaw:
queued for release

arshaw:
released in [3.6.2](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.6.2)

joseluismolinasoria:
Hi.

In version 3.8.0, the same problem occurs again. Fullcalendar doesn't erase the previous "now 

**Html source:**
![now-indicator](https://user-images.githubusercontent.com/3055726/34209222-98b2e5be-e591-11e7-9e13-8fc63def1700.png)

**Render:**
![now-indicator1](https://user-images.githubusercontent.com/3055726/34209223-98d25444-e591-11e7-865d-1452fb3abe27.png)

Thank you.

acerix:
@joseluismolinasoria Would you be able to post a [runnable, stripped-down demonstration of the bug](http://fullcalendar.io/wiki/Reduced-Test-Cases/)? Would really appreciate it because the time saved reproducing will be time spent fixing.

I tried to reproduce the issue, but it's working as expected here:

https://codepen.io/anon/pen/YYGQXq?editors=0010

joseluismolinasoria:
@acerix, I can't isolate the problem in codepen.io. I will keep trying. But I have been able to observe the following behavior:
[...]

PR #3882, opened 2017-10-19
Fix now indicator arrow rendering

dhiemstra:
Fixes issue #3872

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+v3.6.2
+
+Bugfixes:
+- Fixed issue where the now indicator arrow didn't get cleaned up (#3882)
 
 v3.6.1 (2017-10-11)
 -------------------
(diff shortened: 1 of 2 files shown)


neroux:
Nice find, but have you tested it? It still doesnt seem to work I am afraid.

The original code appended the newly created jQuery object to all `.fc-content-skeleton` elements but only added the first **plain** DOM object instance to `nodes`. Your modification now only appends it to the first `.fc-content-skeleton` element but adds **the jQuery object** to `nodes`.

It seems a fix would require both indices, adding the DOM object again.
```js
.appendTo(this.el.find('.fc-content-skeleton')[0])[0]);
```

The question still remains though, why does it suddenly create more than one `.fc-content-skeleton`?

dhiemstra:
Thanks for your quick reply and review @neroux 

Yes, actually this is currently live for a segment of our customers on production ;)

The reason it went wrong before is that it got added to all `.fc-content-skeleton` elements but only the first plain DOM element got added to the `nodes` array. When the remove got called it only removed the first element and left the 2nd.

It works fine now because the jQuery object and the plain DOM object both support `remove()`. However, the jQuery one will remove all of the elements and not just the single plain DOM object.

The addition of another `[0]` in your example will work too, i've changed this. 👍

arshaw:
superseded by https://github.com/fullcalendar/fullcalendar/commit/109f31eb344994e5470034a9c2c0c0978e5a4761

(which is queued for release)

PR #3912, opened 2017-11-05
Replace $.ajax

BrockReece:
Hi Adam,

I thought I would create a PR early so you can keep an eye on what I am working on. To summarise what I have done so far...

- Installed **naugtur/xhr** lib
- Add new gulp task to browserify the xhr lib and add `window.xhr` to fullcallendar.js
- Replace $.ajax in **JsonFeedEventSource**
- Fix tests, build mocks for xhr, remove mockjax

I am guessing we will want to replace the $.ajax based method in **GcalEventSource**, which I have already started but I need to build a better mock xhr method to fix the more comprehensive tests. What are your thoughts so far?

Cheers
Brock

### DIFF
--- a/src/models/event-source/JsonFeedEventSource.js
+++ b/src/models/event-source/JsonFeedEventSource.js
@@ -23,39 +23,39 @@ var JsonFeedEventSource = EventSource.extend({
 		this.calendar.pushLoading();
 
 		return Promise.construct(function(onResolve, onReject) {
-			$.ajax($.extend(
+			xhr($.extend(
 				{}, // destination
 				JsonFeedEventSource.AJAX_DEFAULTS,
 				ajaxSettings,
 				{
 					url: _this.url,
 					data: requestParams,
-					success: function(rawEventDefs) {
-						var callbackRes;
-
-						_this.calendar.popLoading();
-
-						if (rawEventDefs) {
-							callbackRes = applyAll(onSuccess, this, arguments); // redirect `this`
-
-							if ($.isArray(callbackRes)) {
-								rawEventDefs = callbackRes;
-							}
-
-							onResolve(_this.parseEventDefs(rawEventDefs));
-						}
-						else {
-							onReject();
-						}
-					},
-					error: function() {
-						_this.calendar.popLoading();
-
-						applyAll(onError, this, arguments); // redirect `this`
-						onReject();
-					}
 				}
-			));
+			), function(error, rawEventDefs) {
+				if (error) {
+					_this.calendar.popLoading();
+				
+					applyAll(onError, this, arguments); // redirect `this`
+					return onReject();
+				}
+
+				var callbackRes;
+				
+				_this.calendar.popLoading();
+
+				if (rawEventDefs) {
+					callbackRes = applyAll(onSuccess, this, arguments); // redirect `this`
+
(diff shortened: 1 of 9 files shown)


src/models/event-source/JsonFeedEventSource.js
```diff
@@ -23,39 +23,39 @@ var JsonFeedEventSource = EventSource.extend({
...
-					success: function(rawEventDefs) {
-						var callbackRes;
-
-						_this.calendar.popLoading();
-
-						if (rawEventDefs) {
-							callbackRes = applyAll(onSuccess, this, arguments); // redirect `this`
+				}
+			), function(error, rawEventDefs) {
+				if (error) {
```
arshaw: could we do this format...

```js
if (error || !rawEventDefs) {
    // error code
}
else {
    // success code
}
```

tests/legacy/event-feed-param.js
```diff
@@ -10,21 +10,20 @@ describe('event feed params', function() {
 			defaultView: 'month'
 		};
 
-		$.mockjax({
-			url: '*',
-			contentType: 'text/json',
-			responseText: [
+		// mock xhr
+		window.xhr = function(options, callback) {
```
arshaw: this is a clever and lightweight way to intercept the request, but it has it's limitations, namely that a response can't be simulated. also, if we switch xhr libraries in the future, it will break.

also, you need to remember to reassign `window.xhr` to the originally value to truly prevent side effects.

it might be worth looking into this...
https://github.com/jameslnewell/xhr-mock

it has nice `.setup()` and `.teardown()` methods and will be library-independent.

src/models/event-source/JsonFeedEventSource.js
```diff
@@ -23,39 +23,39 @@ var JsonFeedEventSource = EventSource.extend({
...
 
 		return Promise.construct(function(onResolve, onReject) {
-			$.ajax($.extend(
+			xhr($.extend(
 				{}, // destination
 				JsonFeedEventSource.AJAX_DEFAULTS,
 				ajaxSettings,
 				{
 					url: _this.url,
 					data: requestParams,
```
arshaw: is the `data` prop supported by the xhr lib? i thought it was `body`

arshaw: also, i don't think the `cache` prop in `JsonFeedEventSource.AJAX_DEFAULTS` is supported either

arshaw:
also, i think i told u about this over email, but whenever is convenient for you, would be great to switch over to this branch as your base: https://github.com/fullcalendar/fullcalendar/tree/typescript

src/models/event-source/JsonFeedEventSource.js
BrockReece: You are totally right it is `body` and not `data`, it looks like the `body` option doesn't build request params on a GET request either, so will need to handle that too

src/models/event-source/JsonFeedEventSource.js
BrockReece: Yep, no problem

arshaw:
v4 uses superagent, and we might even have it use fetch. either way, using the xhr package is no longer something we need

PR #3915, opened 2017-11-07
Updated agenda-view according to latest test style

jloosli:
@arshaw I tweaked one file to just make sure I was doing it correctly. Please review and let me know if there are things I need to change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ dist
 # npm
 node_modules
 npm-debug.log
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 # npm
 node_modules
 npm-debug.log
+.idea
--- a/tests/legacy/agenda-view.js
+++ b/tests/legacy/agenda-view.js
@@ -1,51 +1,43 @@
+describe('agenda view rendering', function () {
 
-describe('agenda view rendering', function() {
-
-	beforeEach(function() {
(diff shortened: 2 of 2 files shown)


jloosli:
Also, based on the example of Agenda-render.js, jQuery was still left in the tests. Are you wanting to convert to non-jquery dom lookups as this time also, or are you wanting to hold off on that for this first pass?

.gitignore
```diff
@@ -3,4 +3,5 @@ dist
 
 # npm
 node_modules
-npm-debug.log
\ No newline at end of file
+npm-debug.log
+.idea
```
arshaw: could you remove this? leave the .gitignore untouched

arshaw:
hi @jloosli , thanks so much for your help! this looks great.

you can just keep jquery within the tests. the only thing we care about is getting `$().fullCalendar` calls out of there.

i might eventually look for a jasmine plugin that makes this stuff easier, similar to what jasmine-jquery does, for for raw DOM nodes.

jloosli:
@arshaw OK...fixed .gitignore and reverted back to the jQuery version of the test

arshaw:
this might have been auto-closed because of a merge conflict with code i just pushed? i'm not sure

arshaw:
i found a way to get this work into master: https://github.com/fullcalendar/fullcalendar/commit/6924c42fc364aaf7416561876d06441d136e1915

Issue #3311, opened 2016-08-17
eventRender should know if segment is the start/end


svenkatesh-nyros:
How to disable the events fc-title showing its name in the next row first column if the events end date is in next week day.
Ex: start_date: 2016-08-19 end_date: 2016-08-29 the title is displaying in the following dates of week first column like in 21st and 28th day in the calendar.How to stop this.
Please give me suggestion regarding this.Any help would be appreciated.

arshaw:
You can use [eventRender](http://fullcalendar.io/docs/event_rendering/eventRender/) to achieve this but unfortunately there is no clean API for determining if the segment is the starting or ending span of an event, you'll need to query the DOM:

```js
eventRender: function(event, el) {
	if (!el.hasClass('fc-start') && !el.hasClass('fc-end')) {
		el.find('.fc-title').remove();
	}
}
```

leaving this open for discussion on how to improve this.

ChazyTheBest:
Also, if you append an element (like a delete button) inside the `element` (or even deeper `.fc-content`), it will be shown in all weeks.

I think this should be fixed, the default behavior. And the possibility of knowing if segment is start/end could as well be discussed, but it fits better in a PR.

I mean, the default behavior should be to not repeat the element on each week, and if it's long enough it could be displayed in the next weeks, like if the weeks were lines (that could be added as an option).

That's my suggestion. The button, as it's a custom element, we should do the check to show it wherever we want (the logic tell me it should be at the end of the element).

arshaw:
This issue has been addressed in the latest alpha release (4.0.0-alpha.2)
[See the blog post &raquo;](https://fullcalendar.io/blog/2018/10/release-v4-alpha2)

https://fullcalendar.io/docs/v4/eventRender

`isStart` and `isEnd`

arshaw:
I'm closing this issue to signify that it's implemented (since v4-alpha2). I'll comment on this thread again when v4 is officially released.

arshaw:
[V4 was officially released a few days ago!](https://fullcalendar.io/blog/2019/03/v4-officially-released)

PR #3952, opened 2017-11-26
Add segment argument to hook eventRender that allow user to customize event rendering based on information about current rendered segment

w-site:
Add segment argument to hook eventRender that allow user to customize event rendering based on information about current rendered segment.

It will be very helpful to get additional information about segment. For ex. attributes firstRowDayIndex, lastRowDayIndex for customization my event.
For ex. in month view if I know firstRowDayIndex, lastRowDayIndex I can figure out what is the size of current rendered segment of event and based on this information I can attach html that I need, this is one of the case.

I think in any case info about segment will be helpfull for some case.
Thanks.

### DIFF
--- a/src/component/renderers/EventRenderer.ts
+++ b/src/component/renderers/EventRenderer.ts
@@ -155,7 +155,7 @@ export default class EventRenderer {
 					};
 				},
 				filterEl: (seg, el) => {
-					return this.filterEventRenderEl(seg.footprint, el);
+					return this.filterEventRenderEl(seg, el);
 				}
 			});
 		}
@@ -195,7 +195,7 @@ export default class EventRenderer {
 				var el = $(node);
 
 				if (hasEventRenderHandlers) { // optimization
-					el = this.filterEventRenderEl(seg.footprint, el);
+					el = this.filterEventRenderEl(seg, el);
 				}
 
 				if (el) {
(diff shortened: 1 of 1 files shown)


w-site:
About error from travis 
Worker information
hostname: 312a1f82-c019-40b5-b526-4effd3e6d179@1.i-04a0047-production-2-worker-org-ec2.travisci.net
version: v3.4.0 https://github.com/travis-ci/worker/tree/ce0440bc30c289a49a9b0c21e4e1e6f7d7825101
instance: 9019820 travisci/ci-garnet:packer-1503972846 (via amqp)
startup: 853.785803ms
oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"

acerix:
Please submit a ticket under Issues with the feature you are requesting, then link this PR to it.

It looks like you want a way to access the current `segment` from within `eventRender()` , which would also resolve this issue:

https://github.com/fullcalendar/fullcalendar/issues/3311

w-site:
@acerix Thanks for answer.
Yes you are absolutely correct about "It looks like you want a way to access the current segment from within eventRender()", and seems that it resolve https://github.com/fullcalendar/fullcalendar/issues/3311 because segment have options isStart and is End

I created issue as you say above: https://github.com/fullcalendar/fullcalendar/issues/3956

w-site:
This issue has been addressed in the latest alpha release (4.0.0-alpha.2)
See the blog post »

https://fullcalendar.io/docs/v4/eventRender

isStart, isEnd, isMirror at least

PR #3994, opened 2017-12-27
bs.js

sabahuddin:
Translation in Bosnian

### DIFF
--- a/locale/bs.js
+++ b/locale/bs.js
@@ -1,0 +1,41 @@
+import 'moment/locale/ba';
+import * as FullCalendar from 'fullcalendar';
+
+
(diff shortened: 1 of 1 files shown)


acerix:
Great, thank you

locale/bs.js
```diff
@@ -0,0 +1,41 @@
+import 'moment/locale/ba';
```
arshaw: `'moment/locale/ba'` doesn't exist ... it would need to be `'moment/locale/bs'`
(causing the build to fail)

could you change this?

sabahuddin: Done.

arshaw:
released in 3.8.1, thanks

PR #4000, opened 2018-01-04
Converting Constructors in Legacy Tests

acerix:
Starting new batch of updates to legacy tests

https://github.com/fullcalendar/fullcalendar/wiki/Initiatives#how-to-rewrite-legacy-initialization-calls-in-tests


arshaw:
look good, thanks!

PR #4007, opened 2018-01-10
Update locales.html

baijunyao:


acerix:
Thanks for the correction.

This removes a random "q" in locales.html demo HTML.

arshaw:
thanks. i found this recently too and already committed it.

PR #4014, opened 2018-01-14
bower: bump moment to 2.20.1, which includes Chinese fixes

yan12125:
Motivation: Handling of Chinese dates are broken in Nextcloud (https://github.com/nextcloud/calendar/issues/534). It turns out that locale handling was broken in older moment.js versions. This is fixed in https://github.com/moment/moment/pull/3952 and included in moment.js 2.20.0.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -23,7 +23,7 @@
   "copyright": "2017 Adam Shaw",
   "dependencies": {
     "jquery": "2 - 3",
-    "moment": "^2.9.0"
+    "moment": "^2.20.1"
   },
   "main": [
     "dist/fullcalendar.js",


acerix:
Thanks for the PR

arshaw:
thanks, will get this into the next release.

note to self: why isn't build process fetching latest possible versions of dependencies? https://github.com/fullcalendar/fullcalendar/blob/master/bin/build-release.sh#L34

arshaw:
im making a patch-release right now. since this is such a larger version jump for moment, i'll release this with the next minor version release

yan12125:
> i'll release this with the next minor version release

Hello, does that mean this will be included in 3.9? Is there a schedule for that?

arshaw:
released in [v3.9.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v3.9.0)

Issue #2536, opened 2015-08-19
Better accessibility with screen readers


arshaw:
Originally reported on Google Code with ID 2271

```
Inspired by issue 2268. Should research ways to make FullCalendar a better experience
for users with screen readers. Maybe some HTML microformat that does this?
```

Reported by `adamrshaw` on 2014-09-01 22:13:56

Imported with **3** stars.

powrsurg:
Three accessibility issues I can tell you right off the bat:
* On the month view, when you display the "+3 more" when there are too many dates the anchor tag needs an href so that they can tab to the text.
* Continuing from that point, set a tabindex="0" on the close button span on that extended view from the first point so that the user can tab to the close button, and then look for keydown events for the space or enter key.
* Set aria-labels or extra elements with visibly hidden elements to add more text (for example, to make the close button actually read "close").

itayst:
I would like to know what is the status of this issue.
Does full-calendar support today screen readers?

barlowm:
Since I haven't seen any progress on improving the accessibility of this component in meeting with WCAG or Section 508 requirements, does anyone know of any other calendar component that meets accessibility requirements?

ish-joshi:
Here is a way to make the event clicks more accessible. 

Add this option in your fullCalendar configuration.

`eventRender: function (event, element) {
                    element.append("<button class='visuallyhidden' onclick='handleEventClick(" + event + ")' aria-label=' What you want it to say'></button>");
                }`

By adding the invisible button, the screenreader can pick it up. 

Here is the css for .visuallyhidden
`.visuallyhidden {
            border: 0;
            clip: rect(0 0 0 0);
            height: 1px;
            margin: -1px;
            overflow: hidden;
            padding: 0;
            position: absolute;
            width: 1px;
        }`

anupshah:
Hi, I've just recently come across this project and the functionality looks great! Nice one.

On the accessibility side, I wonder if this might help?

If you use aria role=grid, plus various keyboard events to support cursor key navigation, that may give you usability + accessibility enhancements that may also help with screen readers (still needs some additional implementation, mind you)

This has more info: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Grid_Role

Hope that helps?

wright0776:
The main accessibility issue that I am having with FullCalendar is the empty table header on the top left of the table in the day view. Could it be filled with a span that is only viewable to screen readers that says "time" or something like that?

[...]

PR #4023, opened 2018-01-23
Add aria-label to toolbar icon buttons

blissdev:
Very initial work on addressing issues brought forward in #2536. Began by trying to address ARIA labeling of previous/next buttons which only use icons.

**Open Questions**
1. Any advice on the approach or specific implementation details would be appreciated.
2. Are tests desired to check for the attribute? How to test a11y?
2. It doesn't seem there is currently appropriate text for these buttons, which for English would be "Previous" and "Next" and would be needed in all other locales as well, so what is the appropriate way to acquire this text? Can we start with a few common locales?
3. What else is needed to make this addition viable?

Any feedback would be appreciated, thanks!

### DIFF
--- a/src/Toolbar.ts
+++ b/src/Toolbar.ts
@@ -80,6 +80,7 @@ export default class Toolbar {
           let buttonInnerHtml
           let buttonClasses
           let buttonEl
+          let buttonAriaAttr
 
           if (buttonName === 'title') {
             groupChildren = groupChildren.add($('<h2>&nbsp;</h2>')) // we always want it to take up height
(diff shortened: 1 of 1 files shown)


blissdev:
There's a test failure:

```
Expected ({ 0: HTMLNode, length: 1, prevObject: ({ 0: HTMLNode, length: 1 }) }) to have text 'تموز يوليو ٢٠١٦'.
/home/travis/build/fullcalendar/fullcalendar/tmp/compiled-tests.js:13777:39
```

This seems unrelated to my changes, but please let me know if I'm missing something.

acerix:
Thanks for the PR

arshaw:
released. thanks

blissdev:
@arshaw @acerix Hi, I'm a bit confused. This is only the beginning of the work. Currently as implemented this only uses the `buttonName` as the `aria-label`. From what I can tell in English this will be "prev" and "next", where it should really be "previous". Also in some locales this is represented as a symbol which does not provide accessible text from what I can tell. I was hoping to get feedback on how to proceed foward, I don't think it is ready to land yet.

acerix:
Using the button text is better than nothing, and a good start on https://github.com/fullcalendar/fullcalendar/issues/2536.

Other buttons are translated with the [buttonText](https://github.com/fullcalendar/fullcalendar/blob/master/src/options.ts#L51) object, so I suggest making a copy of that eg. `buttonAriaLabels` and using it in the same way to add the proper text.  I agree it should be "previous" instead of "prev" and change based on the selected locale, but not sure that "prev" could just be replaced by "previous" in `buttonText`.

Tests are always nice, but this change is pretty straightforward and can easily be seen in the demos (looking at DOM).  Maybe something like [gulp-accessibility](https://www.npmjs.com/package/gulp-accessibility) could test for a11y in general.

PR #4066, opened 2018-02-19
added DE trad for Year button

Anatakun:
The locales hav not all been updated with the year button traduction. Fixed for the de locale.


acerix:
Thanks

PR #4079, opened 2018-02-25
Update legacy/footer_navigation.js test to new format

NixFifty:
@arshaw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ dist
 # npm
 node_modules
 npm-debug.log
-/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist
 # npm
 node_modules
 npm-debug.log
+/.idea
--- a/tests/automated/legacy/events-gcal.js
+++ b/tests/automated/legacy/events-gcal.js
@@ -23,14 +23,12 @@ describe('Google Calendar plugin', function() {
   var currentWarnArgs
   var oldConsoleWarn
 
+  pushOptions({
+    defaultView: 'month',
+    defaultDate: DEFAULT_MONTH + '-01'
+  })
+
   beforeEach(function() {
-    affix('#cal')
-
-    options = {
-      defaultView: 'month',
-      defaultDate: DEFAULT_MONTH + '-01'
-    }
-
     // Intercept calls to console.warn
     currentWarnArgs = null
     oldConsoleWarn = console.warn
@@ -51,7 +49,7 @@ describe('Google Calendar plugin', function() {
     options.timezone = 'local'
     options.eventAfterAllRender = function() {
       var currentRequest = $.mockjax.unmockedAjaxCalls()[0]
-      var events = $('#cal').fullCalendar('clientEvents')
+      var events = currentCalendar.clientEvents()
       var i
 
       expect(currentRequest.data.timeMin).toEqual(REQUEST_START)
@@ -65,7 +63,7 @@ describe('Google Calendar plugin', function() {
 
       done()
     }
-    $('#cal').fullCalendar(options)
+    initCalendar(options)
   })
 
   it('request/receives correctly when UTC timezone', function(done) {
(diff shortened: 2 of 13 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ dist
 # npm
 node_modules
 npm-debug.log
+/.idea
```
arshaw: can u take this out of the PR?

arshaw:
looks good!

.gitignore
NixFifty: @arshaw I've pushed another commit to revert this

.gitignore
```diff
@@ -6,4 +6,4 @@ dist
 
 # npm
 node_modules
-npm-debug.log
+npm-debug.log
```
arshaw: could you not commit this either?

to prevent from accidentally adding this kind of stuff, or other stuff that might be more harmful, I always use `git add -p` ... to learn more about interactive adding see this: https://nuclearsquid.com/writings/git-add/

tests/automated/legacy/views-specific-options.js
```diff
@@ -1,6 +1,18 @@
+
 describe('view-specific options', function() {
   var options
 
+  pushOptions({
```
arshaw: when you add this `pushOptions` statement, you can then remove the following beforeEach statement that does that did the same logic

.gitignore
NixFifty: whoops, dunno how that happened. 😞

tests/automated/legacy/weekNumberTitle.js
```diff
@@ -23,27 +20,27 @@ describe('weekNumberTitle', function() {
...
-        $('#cal').fullCalendar(options)
+        initCalendar(options);
         expect(getRenderedWeekNumberTitle()).toBe('YO')
       })
 
       it('renders correctly when customized and RTL', function() {
         options.isRTL = true
         options.weekNumberTitle = 'YO'
-        $('#cal').fullCalendar(options)
+        initCalendar(options);
```
arshaw: make sure to run `gulp lint` every so often. you'll find that these semicolons are against the code styleguide.

tests/automated/legacy/firstDay.js
```diff
@@ -35,13 +29,12 @@ describe('First Day', function() {
   })
 
   describe('when setting firstDay to 1', function() {
-    beforeEach(function() {
```
arshaw: i actually want to keep the initialization of `firstDay: 1` in this same spot. it causes the `describe` statement to make sense and allows for future `it` statements within the `describe` block that leverage the same base options.

arshaw: use `pushOptions({ /* ... */ })`

tests/automated/legacy/firstDay.js
```diff
@@ -169,28 +156,30 @@ describe('First Day', function() {
 
   describe('when new firstDay options are set', function() {
     it('should change the first day of week to Monday', function() {
-      $('#cal').fullCalendar({
-        firstDay: 1
-      })
+      var options = {}
+      options.firstDay = 1
+
+      initCalendar(options)
```
arshaw: it much more elegant to write this like:

```js
initCalendar({
  firstDay: 1
})
```

tests/automated/legacy/events-gcal.js
```diff
@@ -51,7 +49,7 @@ describe('Google Calendar plugin', function() {
     options.timezone = 'local'
```
arshaw: failing. because `options` is no longer defined

tests/automated/legacy/hiddenDays.js
```diff
@@ -1,38 +1,31 @@
...
-    beforeEach(function() {
-      $('#cal').fullCalendar({
-        hiddenDays: []
-      })
+    pushOptions({
+      hiddenDays: []
     })
+    
     it('should return 7 days of the week', function() {
       var daysCount = $('.fc-day-header').length
```
arshaw: this test and others like it are failing. because a calendar is not longer being initialized each test

Issue #4122, opened 2018-03-25
GCal: Extended Properties are not passed along


mrnnn:
# About Extended properties
Extended properties allow you to set some meta data on google calendar events. They are not visible but can be set through the Google API. This is often used for meta data, like a tag or category for your event. 

# Problem
It would be super nice to attach a class depending on an extended property, or influence how an event is rendered. However, the gcal plugin doesn't pass the properties along to the fullcalendar event.

# Solution
Just pass along the shared extende properties to the fullcalendar event. The fix is small yet does the trick.

arshaw:
this was merged and released a while ago

PR #4123, opened 2018-03-25
GCal: pass extendedProperties along

mrnnn:
Implements feature request #4122 

# About Extended properties
Extended properties allow you to set some meta data on google calendar events. They are not visible but can be set through the Google API. This is often used for meta data, like a tag or category for your event. 

# Problem
It would be super nice to attach a class depending on an extended property, or influence how an event is rendered. However, the gcal plugin doesn't pass the properties along to the fullcalendar event.

# Solution
Just pass along the shared extende properties to the fullcalendar event. The fix is small yet does the trick.

### DIFF
--- a/plugins/gcal/GcalEventSource.ts
+++ b/plugins/gcal/GcalEventSource.ts
@@ -102,10 +102,10 @@ export default class GcalEventSource extends EventSource {
     if (url && gcalTimezone) {
       url = injectQsComponent(url, 'ctz=' + gcalTimezone)
     }
-    
-    let meta_data = {};
+
+    let metaData = {}
     if (typeof item.extendedProperties === 'object' && typeof item.extendedProperties.shared === 'object') {
-      meta_data = item.extendedProperties.shared;
+      metaData = item.extendedProperties.shared
     }
 
     return {
@@ -116,7 +116,7 @@ export default class GcalEventSource extends EventSource {
       url: url,
       location: item.location,
       description: item.description,
-      meta_data: meta_data
+      metaData: metaData
     }
   }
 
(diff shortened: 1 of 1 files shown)


mrnnn:
Any updates on this? Is this something in the scope of fullcalendar? Or will this never be implemented? @arshaw

arshaw:
thanks @mrnnn ill try to get this into the next release

mrnnn:
Awesome! Thanks!

arshaw:
i've merged this and will release this in the next v3 release really soon (today?). I renamed the property from `metaData` to the more literal `extendedProperties` (because that's what Google calls it).

I've also merged this into the v4 branch, with some modifications of my own. these properties will be accessible via the `extendedProps` hash on each [Event Object](https://fullcalendar.io/docs/v4/event-object).

I did not write any tests for this because it was too much work for me to set up an example google calendar feed and populate these extended props via API calls, but if you could set that up, that'd be great

arshaw:

mrnnn:
Awesome!!! 👍 🎉

magicmb:
+1 for making GCal ExtendedProperties available. 

I am using **v3.10** and hence assume I need to use a property called **extendedProperties** and not extendedProps which presumably is for v4, right? 

Also I am not entirely clear on how to get to all these extended properties. Can't find any detailed docs/examples relating to this.

Read another thread [docs for extended props in v4 #4381](https://github.com/fullcalendar/fullcalendar/issues/4381) from Andrew @ selectsystems and decided to give the "nsfields" code sample there a go using element.popover for my v3 implementation, which  I picked up on in a codpen mentioned in [/docs/event-tooltip-demo](https://fullcalendar.io/docs/event-tooltip-demo) ... but all I get is **undefined**. Pardon my ignorance (see below), but does this mean there are no extended properties for the given event I'm querying?

I found FullCalendar to be quite a learning curve over the past few weeks and I'm actually amazed by what I've accomplished this far, but I really struggle with docs and especially the lack of code samples. This leads me to spend hours on stack overflow/Google searches,  but often even this isn't enough especially when it comes to FullCalendar using GCal eventSources.

Could really do with an example of how to access a GCal extended property in v3.10.

PR #4132, opened 2018-04-02
Update legacy tests

acerix:
More to come

### DIFF
--- a/tests/automated/legacy/header-rendering.js
+++ b/tests/automated/legacy/header-rendering.js
@@ -1,23 +1,18 @@
-
 describe('header rendering', function() {
-
-  beforeEach(function() {
-    affix('#calendar')
-  })
 
   describe('when using default header options', function() {
     it('should have title as default on left', function() {
-      $('#calendar').fullCalendar()
-      expect($('#calendar > .fc-toolbar > .fc-left > *')).toBeMatchedBy('h2')
+      initCalendar()
+      expect($('.fc-toolbar > .fc-left > *', currentCalendar.el)).toBeMatchedBy('h2')
     })
     it('should have empty center', function() {
-      $('#calendar').fullCalendar()
-      var center = $('#calendar > .fc-toolbar > .fc-center')
+      initCalendar()
+      var center = $('.fc-toolbar > .fc-center', currentCalendar.el)
       expect(center).toBeEmpty()
     })
     it('should have right with today|space|left|right', function() {
-      $('#calendar').fullCalendar()
-      var rightChildren = $('#calendar > .fc-toolbar > .fc-right > *')
+      initCalendar()
+      var rightChildren = $('.fc-toolbar > .fc-right > *', currentCalendar.el)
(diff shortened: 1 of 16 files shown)


tests/automated/legacy/header-rendering.js
```diff
@@ -29,92 +24,85 @@ describe('header rendering', function() {
...
         })
       })
     })
   })
 
   describe('when calendar is within a form', function() {
-    beforeEach(function() {
-      $('#calendar').wrap('<form action="http://google.com/"></form>')
-    })
+    var el = $('<div id="calendar"/>')
```
arshaw: this needs to be wrapped with a `beforeEach`. it would also be nice to remove the element in a `afterEach`. but don't worry, i will fix this.

tests/automated/legacy/refetchEvents.js
```diff
@@ -90,7 +85,7 @@ describe('refetchEvents', function() {
     describe('and one event source is asynchronous', function() {
       it('original events remain on the calendar until all events have been refetched', function(done) {
         // set a 100ms timeout on this event source
-        options.eventSources[0].events = function(start, end, timezone, callback) {
+        eventSources[0].events = function(start, end, timezone, callback) {
```
arshaw: because `eventSources` is mutated here, it's best to initialize `eventSources` in a beforeEach. i will make this change

tests/automated/legacy/scroll-state.js
```diff
@@ -1,27 +1,21 @@
...
-      defaultDate: '2015-02-20',
-      contentHeight: 200
-    }
-    affix('#cal')
-    $('#cal').width(800)
+  pushOptions({
+    defaultDate: '2015-02-20',
+    contentHeight: 200
   })
+  var calendarEl = $('<div id="calendar">').width(800).appendTo('body')
```
arshaw: needs to be within a `beforeEach`, or within the `it` block (anything with side-effects should be like this)

tests/automated/legacy/updateEvent.js
```diff
@@ -256,15 +260,17 @@ describe('updateEvent', function() {
       it('should move the end by the delta', function() {
         var event, relatedEvent
 
-        options.events = [
-          { id: '1', start: '2014-05-01T12:00:00', end: '2014-05-01T14:00:00', allDay: false, className: 'mainEvent' },
-          { id: '1', start: '2014-05-10T16:00:00', end: '2014-05-10T19:00:00', allDay: false, className: 'relatedEvent' }
-        ]
-        $('#cal').fullCalendar(options)
+        initCalendar({
+          defaultTimedEventDuration: { hours: 1 },
```
arshaw: i think this `defaultTimedEventDuration` option might be unneeded and accidental?

arshaw:
merged. my fixes: https://github.com/fullcalendar/fullcalendar/commit/c59e5d49daaf1b207d2605b885317c60804348d8

PR #4137, opened 2018-04-06
Update es.js

xchwarze:
Fix names in spanish


acerix:
Thanks for the PR.  Month and day names are not typically not capitalized in Spanish (eg. https://www.thoughtco.com/capitalization-in-spanish-3080293) so I'm wondering the basis for this change.

xchwarze:
Hello Acerix
Thats correct, but when they are the beginning of a sentence they had to be in capital
This is a example of spanish calendar: https://www.calendario-365.es/jpg/calendario-2018.jpg

The only detail I find is that titleFormat would only be correct when it starts with MMMM

arshaw:
fullcalendar only uses the "datepicker" translations in the translation file for the Today/Prev/Next buttons, which are already capitalized.

the capitalization of the months/days you'd like to change are the responsibility of momentjs, which has all lowercase for these.

closing this because we don't want to mutate the moment translations, nor do we want to maintain a separate set of translations.

this won't be an issue in v4 (i hope), which uses the browser's native localization

Issue #2978, opened 2015-12-22
always show more-events-popover with eventLimit 0


pauldemian:
Hey,

I am looking for a way to show the "show more" box (which can be triggered by setting the eventLimit property to a positive number) for all dates, which have events.

I assumed that this could be done by setting the eventLimit to 0, but this is interpreted as "false".
See: http://jsbin.com/cubedi/edit?js,output

Right now I create the missing elements manually, which might cause some problems now or then.

Thanks!

mrhaw:
This would be awesome!

froasiooneloop:
This would be great. It is very common in a month view to have a yes/no event indicator, no matter how many of them you have.

evgeniynickolaev:
Maybe someone knows a workaround for this to make a dirty solution until the issue is closed?

acontreras89:
Indeed there are unnecessary tests for truthiness. See `BasicView` [for an example.](https://github.com/fullcalendar/fullcalendar/blob/5af235874b3fd95b57ec9f632e2e8b9e8c26f6c3/src/basic/BasicView.ts#L179-L181)

Simply using `typeof eventLimit === 'number'` would already be a test for definition. This seems pretty much unintended.

@arshaw would you take a PR on this?

lbrevity:
Any update on this "More available" ?

arshaw:
this is fixed in [the latest v5-beta](https://fullcalendar.io/blog/2020/04/new-ui-features-beta)

acontreras89:
@arshaw should we close #4169 then?

arshaw:
works in v5 with `dayMaxEvents:0`, which has [just been officially released](https://codepen.io/arshaw/pen/PoZmPjV?editable=true&editors=001).

https://codepen.io/arshaw/pen/PoZmPjV?editable=true&editors=001

PR #4169, opened 2018-04-28
Support a value of 0 for eventLimit

acontreras89:
Addresses #2978.

**WIP:**
- [x] Approximate solution
- [x] Get some feedback
- [x] Finish off
- [x] Review specs
- [ ] Update docs?

### DIFF
--- a/src/agenda/AgendaView.ts
+++ b/src/agenda/AgendaView.ts
@@ -13,7 +13,7 @@ import View from '../View'
 import TimeGrid from './TimeGrid'
 import DayGrid from '../basic/DayGrid'
 
-const AGENDA_ALL_DAY_EVENT_LIMIT = 5
+const AGENDA_ALL_DAY_EVENT_LIMIT = 4
 
 let agendaTimeGridMethods
 let agendaDayGridMethods
--- a/src/agenda/AgendaView.ts
+++ b/src/agenda/AgendaView.ts
@@ -201,7 +201,7 @@ export default class AgendaView extends View {
       if (eventLimit && typeof eventLimit !== 'number') {
         eventLimit = AGENDA_ALL_DAY_EVENT_LIMIT // make sure "auto" goes to a real number
       }
-      if (eventLimit) {
+      if (eventLimit !== false) {
         this.dayGrid.limitRows(eventLimit)
       }
     }
--- a/src/basic/BasicView.ts
+++ b/src/basic/BasicView.ts
@@ -176,7 +176,7 @@ export default class BasicView extends View {
     this.dayGrid.removeSegPopover() // kill the "more" popover if displayed
 
     // is the event limit a constant level number?
-    if (eventLimit && typeof eventLimit === 'number') {
+    if (typeof eventLimit === 'number') {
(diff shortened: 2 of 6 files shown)


acontreras89:
This approximation changes how `eventLimit` works a little. For instance, if we have three events in one day and a limit of 2:

* Before, only one event would be displayed and a +2 badge would be added
* Now, two events are displayed and a +1 badge is added

This was the quickest approach to be able to handle 0 as a valid value for the limit.

I've had to update current tests and the internal default value for `AgendaView`.

src/agenda/AgendaView.ts
```diff
@@ -13,7 +13,7 @@ import View from '../View'
 import TimeGrid from './TimeGrid'
 import DayGrid from '../basic/DayGrid'
 
-const AGENDA_ALL_DAY_EVENT_LIMIT = 5
+const AGENDA_ALL_DAY_EVENT_LIMIT = 4
```
arshaw: it seems that in many places in this PR, the event limit numeric value is now being tested for one-less that it was previously being tested for. did the meaning of "event limit" change or something? trying to understand at a high level

acontreras89: If I recall correctly, the old behavior was to display a maximum of `eventLimit` items. For instance, with a limit of 3 the user could see:

* One up to three events, or
* Two events and a *+N* badge

This has changed now, and the limit is inclusive. The user will see three events and the badge.

acontreras89:
There's a bug here that I have already solved in my fork (https://github.com/acontreras89/fullcalendar/commit/5a21d839ffccdea939f891fcb395e727b7da49cc). I'll update the PR when I have the time.

acontreras89:
Updated.

Can anyone give me a hint as to what may be causing CI to fail?

src/agenda/AgendaView.ts
acontreras89: Sorry I missed this comment.

With an `eventLimit` of `3` and 4 events, the user would see three of them and a badge (+1.) Is that what you were asking?

acontreras89:
@arshaw do we need anything else?

arshaw:
I'm going through old PRs right now.

I will accept this change but only if it is isolated to "always show more-events-popover with eventLimit 0"

As I noted a while back, this also changed the definition of `eventLimit` to be one-less. I believe these other issues match this request as well: #3129, #3035. I need to think about that API change more.

acontreras89:
> I will accept this change but only if it is isolated to "always show more-events-popover with eventLimit 0"

There's no easy way to isolate this. Code has things like `td = cellMatrix[levelLimit - 1][seg.leftCol]`, which will crash if we feed a value of 0.

This is because `eventLimit` is conceptually (and semantically) wrong. For a value of _N_ and more than _N_ events, the _Nth_ event is being hidden. But hidding the 0th event doesn't make any sense, does it?

arshaw:
closing per my comment here:https://github.com/fullcalendar/fullcalendar/issues/2978#issuecomment-647795671

thanks @acontreras89

PR #4183, opened 2018-05-13
Fix mistranslation in Japanese

hissy:
Original translate string means "Do not display events".

### DIFF
--- a/locale/ja.js
+++ b/locale/ja.js
@@ -35,5 +35,5 @@ FullCalendar.locale("ja", {
   eventLimitText: function(n) {
     return "他 " + n + " 件";
   },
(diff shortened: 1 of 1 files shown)


arshaw:
thanks!

arshaw:
I already merged #4329 which does the same thing. not sure which one is better because I don't speak Japanese!

hissy:
"表示する予定はありません" and "表示するイベントがありません" are almost same sentence, the only defference is the word "予定" and "イベント". 
"イベント" is the completely same meaning with the English word "Event". 
"予定" is a bit different word, "Plan" is the almost same meaning.

PR #4184, opened 2018-05-14
first few

jstacoder:
pulled dom stuff out of event-drag/repeating and event-data/refetchEvents

### DIFF
--- a/tests/automated/event-data/refetchEvents.js
+++ b/tests/automated/event-data/refetchEvents.js
@@ -1,8 +1,10 @@
+import { findFcEvent, findScroller, createCalElement } from '../lib/events'
 
 describe('refetchEvents', function() {
 
   it('retains scroll when in month view', function() {
-    var el = $('<div id="calendar" style="width:300px"/>').appendTo('body')
+    // var el = $('<div id="calendar" style="width:300px"/>').appendTo('body')    
+    var el = createCalElement({style:"width:300px"}).appendTo('body')
     var scrollEl
     var scrollTop
 
@@ -21,9 +23,9 @@ describe('refetchEvents', function() {
       ]
     }, el)
 
-    expect($('.fc-event').length).toBe(8)
+    expect(findFcEvent().length).toBe(8)
 
-    scrollEl = el.find('.fc-scroller')
+    scrollEl = findScroller(el)
     scrollEl.scrollTop(1000)
     scrollTop = scrollEl.scrollTop()
 
@@ -31,7 +33,7 @@ describe('refetchEvents', function() {
     expect(scrollTop).toBeGreaterThan(10)
 
     currentCalendar.refetchEvents()
-    expect($('.fc-event').length).toBe(8)
+    expect(findFcEvent().length).toBe(8)
     expect(scrollEl.scrollTop()).toBe(scrollTop)
   })
 })
--- a/tests/automated/event-data/refetchEvents.js
+++ b/tests/automated/event-data/refetchEvents.js
@@ -3,7 +3,6 @@ import { findFcEvent, findScroller, createCalElement } from '../lib/events'
 describe('refetchEvents', function() {
 
   it('retains scroll when in month view', function() {
-    // var el = $('<div id="calendar" style="width:300px"/>').appendTo('body')    
     var el = createCalElement({style:"width:300px"}).appendTo('body')
     var scrollEl
     var scrollTop
--- a/tests/automated/event-drag/repeating.js
+++ b/tests/automated/event-drag/repeating.js
@@ -40,9 +40,11 @@ describe('event dragging on repeating events', function() {
       eventDragStart: function() {
         setTimeout(function() { // try go execute DURING the drag
           expect(
-            $('.fc-event:visible').filter(function(i, node) {
-              return $(node).css('visibility') !== 'hidden'
-            }).length
+            TimeGridEventDragUtils.filterEl('.fc-event:visible', 
+              function(i, node) {
+                return $(node).css('visibility') !== 'hidden'
+              }
+            ).length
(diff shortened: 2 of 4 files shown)


jstacoder:
going to close to figure out whats going on with travis

tests/automated/event-data/refetchEvents.js
```diff
@@ -21,17 +22,17 @@ describe('refetchEvents', function() {
       ]
     }, el)
 
-    expect($('.fc-event').length).toBe(8)
+    expect(findFcEvent().length).toBe(8)
 
-    scrollEl = el.find('.fc-scroller')
+    scrollEl = findScroller(el)
```
arshaw: we are attempting to grab the area of month view that becomes scrollable when there is not enough vertical space to fit all the day rows, so i'd like to make this part of the month view utilities. i'm not sure we have a file for that yet. maybe create a `MonthViewUtils` file? Also, i'd like to rename this to `getScrollerEl` since it returns an el. And I like 'get' better than 'find' because 'find' was simply chosen because it mirrors the jquery method name. also, we can have all of these utilities operate off of the global `currentCalendar` . So, something like this:

in `MonthViewUtils.js`:

```js
export function getScrollerEl() {
   return $(currentCalendar.el).find('.fc-scroller');
}
```

tests/automated/event-drag/repeating.js
```diff
@@ -53,15 +55,14 @@ describe('event dragging on repeating events', function() {
       }
     })
 
-    $('.fc-event:first').simulate('drag', {
+    TimeGridEventDragUtils.simulateDrag('.fc-event:first', {
```
arshaw: you're adding another layer here, but it's not increasing the abstraction or making it more readable, so it's unnecessary.

arshaw: also, why did you decide to put all these tests in `TimeGridEventDragUtils` ? all the tests are running month view, which doesn't have a time grid. Remember, a time grid is the thing with the vertical time axis?

tests/automated/event-drag/repeating.js
```diff
@@ -76,8 +77,9 @@ describe('event dragging on repeating events', function() {
       eventDragStart: function() {
         setTimeout(function() { // try go execute DURING the drag
           expect(
-            $('.fc-event:visible').filter(function(i, node) {
-              return $(node).css('visibility') !== 'hidden'
+            TimeGridEventDragUtils.filterEl('.fc-event:visible', 
+              function(i, node) {
+                return $(node).css('visibility') !== 'hidden'
```
arshaw: similar to my previous complaint, you're adding abstraction here, but to no benefit. a better way to do it is like:

```js
getVisibleEventEls().filter(function() {
    // stuff
})
```

tests/automated/lib/events.js
```diff
@@ -0,0 +1,11 @@
+export function findFcEvent(){
+  return $(".fc-event")
+}
+
+export function findScroller(el){
```
arshaw: you're going to be moving 2 out of the 3 of these function to different files, but i still want to ask, why are these functions within the `events` file? `findScroller` and `createCalElement` having nothing to do with events.

tests/automated/event-drag/repeating.js
arshaw: so, you'd probably want to do something like this instead

```js
getEventEls().first().simulateDrag({
```

the goal is not to completely removing jquery from the tests. the goal is to remove classnames from the tests.

tests/automated/event-data/refetchEvents.js
```diff
@@ -1,8 +1,9 @@
+import { findFcEvent, findScroller, createCalElement } from '../lib/events'
 
 describe('refetchEvents', function() {
 
   it('retains scroll when in month view', function() {
-    var el = $('<div id="calendar" style="width:300px"/>').appendTo('body')
```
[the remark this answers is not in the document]
jstacoder: Good point, I was mainly focusing initially on the grep result. I see what you mean here.

tests/automated/event-data/refetchEvents.js
```diff
@@ -21,17 +22,17 @@ describe('refetchEvents', function() {
       ]
     }, el)
 
-    expect($('.fc-event').length).toBe(8)
+    expect(findFcEvent().length).toBe(8)
```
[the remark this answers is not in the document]
jstacoder: I completely agree

tests/automated/event-data/refetchEvents.js
jstacoder: Good point on placement & naming, I was focusing too much on the test file names / original method names

tests/automated/event-drag/repeating.js
jstacoder: I really like that idea, will get the syntax more like that

tests/automated/lib/events.js
jstacoder: Just initial growing pains / tired after long night of getting tests to pass, pretty sure. So where would we want the “most general” utils that might be used in any file? (I believe that was my original idea for events.js, although I see how that wasn’t ideal, or useful) but stuff not necessarily confined to an individual view?

arshaw: no worries. here are some files we currently have that already express this:

- `event-render/TimeGridEventRenderUtils.js` - event rendering on the vertical time axis
- `event-render/DayGridEventRenderUtils.js` (not yet created) - event rendering on month view days or the all-day in agenda view
- `event-render/EventRenderUtils.js` - event rendering utils that apply to *any* view

try your best to fit the utils into files that express what they do. where they are called from should be irrelevant. also, avoid really vague or "general" modules... there's always some sort of way to make a taxonomy of these utils.

PR #4186, opened 2018-05-18
Jquery removal

jstacoder:
i put all of the functions in monthViewUtils, but if they should go anywhere else just let me know


arshaw:
sounds great, good to keep committing to this branch, with same pr

arshaw:
hey kyle, i haven't given this a thorough look yet, but it looks like you're headed in the right direction. can you remember to run `gulp lint` every so often to check styleguide violations? i see some (of course i never mentioned this to you, so no worries). if you see violations could you fix them? thought i'd bring it up before you get much further

jstacoder:
Yea no problem, I’ll make sure to do that.

tests/automated/lib/MonthViewUtils.js
```diff
@@ -0,0 +1,50 @@
+const EVENT_CLASS = 'fc-event'
+const SCROLLER_CLASS = 'fc-scroller'
+const TITLE_CLASS = 'fc-title'
+const RESIZER_CLASS = 'fc-resizer'
+const TIME_CLASS = 'fc-time'
+
+export function getEventElTime(el){
```
arshaw: to prevent confusion of whether this function returns a millisecond time, or an object that represents the time (like a moment.Duration) **VERSUS** actual text, could you rename it to `getEventElTimeText` ?

tests/automated/lib/MonthViewUtils.js
```diff
@@ -0,0 +1,50 @@
...
+
+export function getVisibleEventEls(){
+  return $(`.${EVENT_CLASS}:visible`)
+}
+
+export function getScrollerEl(){
+  return $(currentCalendar.el).find(`.${SCROLLER_CLASS}`);
+}
+
+export function getEventEls(eventClassName=EVENT_CLASS){
```
arshaw: is the `eventClassName` argument being used? i don't see any callers using it. the files that call these utils, I don't want them to think about classNames, so I see no reason for this

tests/automated/lib/MonthViewUtils.js
```diff
@@ -0,0 +1,50 @@
...
+  return $(`.${eventClassName}`)
+}
+
+export function getSingleEventEl(eventEl){
```
arshaw: is this being called from anywhere? regardless, i don't like having this AND `getFirstEventEl` ... it's not clear what the difference is. after reading the source, i see the different is the `expect` statement, but i'd prefer the caller file to deal w/ the `expect` statement (in other words, don't have this utility file, or any utility file, do expect statements). solely use `getFirstEventEl`, and if there is no first el, have it return null, and have the caller expect it to be an object

tests/automated/lib/MonthViewUtils.js
```diff
@@ -0,0 +1,50 @@
...
+
+export function getLastEventEl(eventClassName){
+  return getSingleEventEl(getEventEls(eventClassName).last())
+}
+
+export function getDayTdEls(date){
+  return $(`td[data-date="${date}"]`)
+}
+
+export function getEventTitleEl(eventEl) {
```
arshaw: When I see the word `Event` in this title, I expect it to accept an event object, or an event ID, but really it expects an element. If one of my function or variables is about an element, I nearly always name it with "el". Could we do that here, and with `getEventResizerEl` as well? i know it gets a little verbose, but i'd prefer it

So, rename to `getEventElTitleEl` and `getEventElResizerEl`

tests/automated/lib/time-grid.js
```diff
@@ -129,13 +129,26 @@ export function getTimeGridTop(targetTime) {
     Math.min(1, (targetTime - slotTime) / slotMsDuration)) // don't go past end of last slot
 }
 
+export function getElTime(el) {
```
arshaw: this method is only for event elements, not any type of element in general, so can we name it so?

rename to `getEventElTimeEl`

tests/automated/lib/time-grid.js
```diff
@@ -129,13 +129,26 @@ export function getTimeGridTop(targetTime) {
...
+  return el.find('.fc-time')
+}
 
 export function getTimeGridDayEls(date) {
   date = FullCalendar.moment.parseZone(date)
 
   return $('.fc-time-grid .fc-day[data-date="' + date.format('YYYY-MM-DD') + '"]')
 }
 
+export function getAllTimeGridSlotEls() {
```
arshaw: no need for the word "all" in this name. The plural of "els" implies it in a way.

rename to `getTimeGridSlotEls`

tests/automated/lib/time-grid.js
```diff
@@ -129,13 +129,26 @@ export function getTimeGridTop(targetTime) {
...
+  return  $('.fc-time-grid .fc-slats tr[data-time]')
+}
+
+
+export function getTimeGridSlotElByIndex(index) {
```
arshaw: i like this name 👍

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -1,7 +1,7 @@
 
 export function getSingleDayEl(date) {
```
arshaw: i see that various callers of this function sometimes specify a `bg` argument, but i don't see it here. what's the deal?

tests/automated/lib/MonthViewUtils.js
```diff
@@ -0,0 +1,50 @@
+const EVENT_CLASS = 'fc-event'
```
arshaw: this comment concerns the `MonthViewUtils.js` as a whole...

It's great that this file is serving as the primary container for all new utilities, and I believe I encouraged you do begin doing this, however, I want to change our approach a bit. I want to put these utils across different files. This should probably be done as the last step of this PR review, after you've made all other changes.

The event-element-related utils in here (`getVisibleEventEls`, `getEventEls`, `getFirstEventEl`, `getLastEventEl`, `getEventElTitleEl`, `getEventResizerEl`), the way I wrote the original statements that they are derived from, they actually query for ANY event elements on the page. They don't discriminate between view (like month vs agenda) or "grid" (time-grid vs day-grid). This is BAD because, most of the time, we want to write tests for the isolated subcomponents, aka the "grids". However, if we want to write a test for a time-grid, and a day-grid is also on the page, and it also has events on it, we'll accidentally be querying for event elements that are outside of the time-grid we're concerned about!

I admit this is bad, but it's the trend I originally set when writing the tests, so I want to keep this for now. However, these types of event-element-related utilities that apply to ALL event elements on the page, I'd like to isolate them in a different file. **that files is: `event-render/EventRenderUtils.js`**

Next, aside from the event-element-related utils in this file, there are other utils that seem to apply to "month" view by nature of them being in this file. `getDayEls`, `getDayTdEls`, and `getScrollerEl`. However, the first two (`getDayEls`, `getDayTdEls`) are better attributed to day-grid. Remember how day-grid is a table of days cells that is X columns wide and X rows tall? and it lives in the all-day section of the Agenda views as well? **So, we should move `getDayEls` and `getDayTdEls` into the file `view-render/DayGridRenderUtils.js`**

The only utility function that should actually *stay* in this file is `getScrollerEl` because it ONLY applies to month view in this case. It is a container element that holds the day-grid and if the day-grid is too tall, it allows the user to scroll. This is related to month-view because it is specific to HOW the month view houses the day-grid within it. Yes, the agenda-views also have scroller els, but they do different things.

tests/automated/lib/MonthViewUtils.js
jstacoder: cool, i had a similar thought, wasnt sure.

tests/automated/lib/MonthViewUtils.js
jstacoder: there was a spot where the test was defined in a function and the class was being passed, ill add a comment at that spot.

tests/automated/lib/MonthViewUtils.js
jstacoder: there are spots asking for `:last` and ':first` so getSingleEventEl was supposed to do the check for one element, its called by getFirst and getLast, but i can pull it out, i only added it because i saw that each time a single element was grabbed it was expected to be a single element, so i thought it would reduce lines of tests, while not actually reducing the tests ran, but yea i can take it out and add the original expects back.

tests/automated/lib/MonthViewUtils.js
jstacoder: thats funny, i originally named it that, but thought it sounded weird, ok cool ill change it.

tests/automated/lib/time-grid.js
jstacoder: no problem

tests/automated/lib/time-grid.js
jstacoder: ok cool,

tests/automated/lib/dnd-resize-utils.js
```diff
@@ -1,5 +1,10 @@
 
 // this function has been mangled to work with external jqui draggables as well
+import {getElTime, getTimeGridDayEls, getTimeGridSlotElByIndex} from "./time-grid";
+import {getEventResizerEl, getEventTitleEl, getFirstEventEl, getLastEventEl} from "./MonthViewUtils";
+import {getSingleDayEl} from "../view-render/DayGridRenderUtils";
+
+
 export function testEventDrag(options, dropDate, expectSuccess, callback, eventClassName) {
```
jstacoder: this is where the eventClassName is passed dynamiclly

tests/automated/lib/dnd-resize-utils.js
```diff
@@ -19,19 +24,17 @@ export function testEventDrag(options, dropDate, expectSuccess, callback, eventC
     eventsRendered = true
 
     dropDate = calendar.moment(dropDate)
-    eventEl = $('.' + (eventClassName || 'fc-event') + ':first')
-    expect(eventEl.length).toBe(1)
+    eventEl = getFirstEventEl(eventClassName)
```
jstacoder: and here is where i added it for, it was originally only defaulting to `.fc-event` so i wasnt sure if i could be certain it always would be.  but if this is always `.fc-event` then taking it out as a parameter should be fine

tests/automated/view-render/DayGridRenderUtils.js
jstacoder: i noticed some had a `.bg-event` class and some didnt, i removed it because i thought it was making the tests fail, but it wasnt, i was going to add that back in

tests/automated/lib/dnd-resize-utils.js
jstacoder: i mean, this spot is why i added the "eventClassName" argument to `getFirstEventEl`

tests/automated/event-data/refetchEvents.js
```diff
@@ -1,3 +1,5 @@
+import { getScrollerEl} from '../lib/MonthViewUtils'
+import {getEventEls} from '../event-render/EventRenderUtils'
```
arshaw: i would expect this to be a linting problem. if not, i'd still like to have inner spaces like `{ getEventEls }`

tests/automated/event-drag/repeating.js
```diff
@@ -1,4 +1,5 @@
 import * as TimeGridEventDragUtils from './TimeGridEventDragUtils'
+import {getEventEls, getVisibleEventEls} from '../event-render/EventRenderUtils'
```
arshaw: same

tests/automated/event-drag/repeating.js
```diff
@@ -89,7 +90,7 @@ describe('event dragging on repeating events', function() {
       }
     })
 
-    $('.fc-event:first').simulate('drag', {
+    getEventEls().first().simulate('drag', {
```
arshaw: this looks like an opportunity to use `getFirstEventEl()`

tests/automated/view-render/DayGridRenderUtils.js
arshaw: the day grid `<td>`'s will ALWAYS be within an .fc-bg div, so no need to make it an option. the times where you saw a selector without the .fc-bg part will work just as well with the .fc-bg specified, i'm nearly sure

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -12,3 +12,11 @@ export function getDisabledEl(i) {
   expect(el).toHaveLength(1)
   return el
 }
+
+export function getDayEls() {
+  return $('.fc-day-header[data-date]')
+}
+
+export function getDayTdEls(date) {
```
arshaw: this function will do the same thing as `getSingleDayEl`, so let's just use that one instead

tests/automated/view-render/DayGridRenderUtils.js
arshaw: could you rename this to `getDayEl`. also, move the `expect` out of here and have the caller always be responsible for checking?

tests/automated/lib/time-grid.js
```diff
@@ -129,15 +129,26 @@ export function getTimeGridTop(targetTime) {
...
+  return $('.fc-time-grid .fc-slats tr[data-time]')
+}
+
+
+export function getTimeGridSlotElByIndex(index) {
+  return $(`.fc-slats tr:eq(${index})`)
+}
 
-export function getTimeGridSlotEls(timeDuration) {
+export function getTimeGridDurationSlotEls(timeDuration) {
```
arshaw: could you rename this to `getSlotElByTime` ... it only returns one slot el for the given time.

tests/automated/lib/time-grid.js
```diff
@@ -129,15 +129,26 @@ export function getTimeGridTop(targetTime) {
...
+  return el.find('.fc-time')
+}
 
 export function getTimeGridDayEls(date) {
   date = FullCalendar.moment.parseZone(date)
 
   return $('.fc-time-grid .fc-day[data-date="' + date.format('YYYY-MM-DD') + '"]')
 }
 
+export function getTimeGridSlotEls() {
```
arshaw: for this func as well as `getTimeGridSlotElByIndex` ... we've developed a convention of not including the grid-name in the method name. in order to stay consistent, let's remove it from here and simply call these methods `getSlotEls` and `getSlotElByIndex`

tests/automated/view-render/DayGridRenderUtils.js
jstacoder: by   
>move the expect out of here and have the caller always be responsible for checking   
  
do you mean to add that expect after places where that function is called? or just from this point on have the caller be responsible?

tests/automated/event-data/refetchEvents.js
jstacoder: yep, defiantly an oversight

tests/automated/event-drag/repeating.js
jstacoder: same

tests/automated/event-drag/repeating.js
jstacoder: my thoughts exactly

tests/automated/view-render/DayGridRenderUtils.js
jstacoder: yep, exact same thing, good idea

tests/automated/lib/time-grid.js
jstacoder: makes sense, i wasnt really sure

tests/automated/lib/time-grid.js
jstacoder: cool, i like the smaller names

jstacoder: should `getSlotEls`  have the `.fc-time-grid` selector removed since we dont want it specific to any grid?

tests/automated/view-render/DayGridRenderUtils.js
jstacoder: uh oh, the selector used in `getDayTdEls` does get the same elements as the selector in `getDayEl`, but it also gets the part of the day elements that has the day number, `.fc-day-top` so i need to keep it for now, because its needed in dayNumbers to pass the tests

tests/automated/lib/time-grid.js
```diff
@@ -129,15 +129,26 @@ export function getTimeGridTop(targetTime) {
     Math.min(1, (targetTime - slotTime) / slotMsDuration)) // don't go past end of last slot
 }
 
+export function getEventElTimeEl(el) {
```
arshaw: wouldn't this utility be best in the general events utility file? not the time-grid file

tests/automated/view-render/DayGridRenderUtils.js
arshaw: ah, ok. if that's the case, and this util is only every used to get the day number text, can we rework it to simply return the day-number text for the given date? it would be best to make the selector more specific to this purpose (by adding `.fc-day-top`). And it would return the `.text()`

tests/automated/lib/dnd-resize-utils.js
```diff
@@ -19,19 +28,19 @@ export function testEventDrag(options, dropDate, expectSuccess, callback, eventC
     eventsRendered = true
 
     dropDate = calendar.moment(dropDate)
-    eventEl = $('.' + (eventClassName || 'fc-event') + ':first')
+    eventEl = getFirstEventEl(eventClassName)
```
arshaw: instead of making this `eventClassName` argument part of `getFirstEventEl` and `getLastEventEl`'s API, could we have the caller be in charge of it. So, something like this:

```js
eventEl = eventClassName ? $('.' + eventClassName  + ':first') : getFirstEventEl()
```

tests/automated/lib/time-grid.js
jstacoder: I thought it belonged here because it’s grabbing a time, but I just realized there are places outside of time-grids that have times. Good point.

tests/automated/lib/dnd-resize-utils.js
jstacoder: Ok on it.

jstacoder: when  i added `.fc-day-top`  and return `.text()` when i run the tests i get the odd error: 
Expected '١' to contain text '١'.

jstacoder: looks like it works if i change the test to `expect().toEqual()`

tests/automated/lib/dnd-resize-utils.js
```diff
@@ -102,17 +112,17 @@ export function testEventResize(options, resizeDate, expectSuccess, callback, ev
     eventsRendered = true
 
     resizeDate = calendar.moment(resizeDate)
-    eventEl = $('.' + (eventClassName || 'fc-event') + ':last')
-    dragEl = eventEl.find('.fc-resizer')
+    eventEl = eventClassName ? $(`.${eventClassName}:first`) : getLastEventEl(eventClassName)
```
arshaw: missed one thing: the param being given to `getLastEventEl` is unnecessary now

PR #4231, opened 2018-06-30
V4

jstacoder:
here is a little one to get my tests passing, and to start using the new functions in the legacy folder.


tests/automated/legacy/Agenda-render.js
```diff
@@ -11,8 +13,8 @@ describe('Agenda view rendering', function() {
 
     it('renders the axis on the left', function() {
       initCalendar()
-      var header = $('.fc-view > table > thead')
-      var firstSlat = $('.fc-slats tr:first')
+      var header = getHeaderEl()
+      var firstSlat = getSlotEls().first()
       expect(header.find('.fc-axis')).toBeLeftOf(header.find('.fc-day-header:first'))
```
jstacoder: for spots like this, where we are doing a bunch of expects on elements calling find on classes, and what not, like these 3 lines 18-20, should i make functions for these? if so maybe something that returns `$('.fc-day-grid') ` that the others can build from?

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -1,28 +1,34 @@
...
   expect(el).toHaveLength(1)
   return el
 }
 
 export function getDayEls() {
-  return $('.fc-day-header')
+  return $(`.${DAY_HEADER_CLASS}`)
 }
 
 export function getDayElTopElText(date) {
```
arshaw: however, you rename `getDayEls`, you'll want to appropriately rename this as well

tests/automated/legacy/allDaySlot.js
```diff
@@ -8,7 +10,7 @@ describe('allDaySlots', function() {
           defaultView: 'agendaWeek'
         }
         initCalendar(options)
-        var allDaySlotCount = $('.fc-day-grid').length
+        var allDaySlotCount = $(`.${DAY_GRID_CLASS}`).length
```
arshaw: use a util function for these

arshaw: the utility you'd create would need to be for AGENDA-view only because agenda view is the only view that has both a time grid AND an all-day row (which is day-grid). you'd want to test that a day-grid is within an agenda view. you'd want to put this util into a file similar to MonthViewUtils, but instead calls it AgendaViewUtils

tests/automated/legacy/agenda-view.js
```diff
@@ -9,15 +15,15 @@ describe('agenda view rendering', function() {
...
-      var headers = $('.fc-view > table > thead th')
-      expect(headers[0]).toHaveClass('fc-axis')
-      expect(headers[1]).toHaveClass('fc-sun')
-      expect(headers[2]).toHaveClass('fc-mon')
-      expect(headers[3]).toHaveClass('fc-tue')
-      expect(headers[4]).toHaveClass('fc-wed')
-      expect(headers[5]).toHaveClass('fc-thu')
-      expect(headers[6]).toHaveClass('fc-fri')
-      expect(headers[7]).toHaveClass('fc-sat')
+      var headers = getHeaderEl().find('th')
```
arshaw: not always guaranteed to be a `th` ... can you make a util function that returns each of the day-of-week header els in one go?

tests/automated/legacy/Agenda-render.js
```diff
@@ -11,11 +14,11 @@ describe('Agenda view rendering', function() {
...
     it('renders the axis on the left', function() {
       initCalendar()
-      var header = $('.fc-view > table > thead')
-      var firstSlat = $('.fc-slats tr:first')
-      expect(header.find('.fc-axis')).toBeLeftOf(header.find('.fc-day-header:first'))
-      expect($('.fc-day-grid .fc-axis')).toBeLeftOf($('.fc-day-grid .fc-day:first'))
-      expect(firstSlat.find('.fc-axis')).toBeLeftOf(firstSlat.find('td:not(.fc-axis)'))
+      var header = getHeaderEl()
+      var firstSlat = getSlotEls().first()
+      expect(header.find(`.${AXIS_CLASS}`)).toBeLeftOf(getFirstDayEl())
```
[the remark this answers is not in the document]
jstacoder: yea, good point, i see what you mean

tests/automated/legacy/agenda-view.js
jstacoder: by not guaranteed to be `th` do you mean the util function should also support the list-views also? becasue im looking at the different views and im not sure where to go to make one function for the 2. should i check the current view in the function to tell it which query string to use?

tests/automated/legacy/allDaySlot.js
jstacoder: sounds good.

tests/automated/lib/constants.js
```diff
@@ -0,0 +1,24 @@
+export const EVENT_CLASS = 'fc-event'
```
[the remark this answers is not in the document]
jstacoder: i like that idea, on it

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -1,28 +1,34 @@
...
 
 export function getDayEl(date) {
   if (typeof date === 'string') {
     date = new Date(date)
   }
-  return $('.fc-day-grid .fc-bg .fc-day[data-date="' + formatIsoDay(date) + '"]')
+  return $(`.${DAY_GRID_CLASS} .${BG_CLASS} .${DAY_CLASS}[data-date="${formatIsoDay(date)}"]`)
 }
 
 export function getDisabledEl(i) {
```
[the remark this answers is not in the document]
arshaw: oh ok i didnt even realize it had to do with "day" elements, since that's not in the name. can you change it to `getDisabledDayElByIndex` ?

tests/automated/legacy/agenda-view.js
arshaw: no, im not thinking about list view, im just saying that `getHeaderEl` returns the wrapper element for the header of month/basic/agenda views, and let's also have a util that returns the individual cells in the header. so instead of `getHeaderEl().find('th')` we'd do `getHeaderCellEls()`

tests/automated/legacy/displayEventEnd.js
```diff
@@ -27,7 +29,7 @@ describe('displayEventEnd', function() {
               allDay: true
             } ],
             eventAfterAllRender: function() {
-              expect($('.fc-event .fc-time').length).toBe(0)
+              expect(getEventElTimeEl(getEventEls()).length).toBe(0)
```
arshaw: the `getEventEls` implies that multiple event elements might be returned, but `getEventElTimeEl` only accepts one ... use a different util method that only returns on event element?

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -1,38 +1,99 @@
...
+
+export function getDayGridDayEl(){
+    return $(`.${DAY_GRID_CLASS} .${DAY_CLASS}`)
+}
+
+export function getFirstDayGridDayEl(){
+  return getDayGridDayEl().first()
+}
+
+export function getGetMoreEl(){
```
arshaw: is this method name a typo?

tests/automated/legacy/displayEventEnd.js
jstacoder: i was just trying to reuse the functions we are writing, since this combination gives the desired result, but if its better to be readable than dry, i can make another function, i guess `getEventEl` ?

tests/automated/view-render/DayGridRenderUtils.js
jstacoder: yea, ill fix that

tests/automated/legacy/agenda-view.js
jstacoder: i guess what i mean is if it isnt always going to be a `th` what should `getHeaderCellEls` return?

tests/automated/event-render/EventRenderUtils.js
```diff
@@ -1,25 +1,37 @@
...
+const START_CLASS = 'fc-start'
+const END_CLASS = 'fc-end'
+const DAY_CLASS = 'fc-day'
+const EVENT_DOT_CLASS = 'fc-event-dot'
+
+export function getEventElDotEl(el) {
+  return el.find(`.${EVENT_DOT_CLASS}`)
+}
+
+export function replaceEventElDotElWithEl(eventEl, newDotEl){
```
arshaw: you have a lot of this type of linter problem. `){`

tests/automated/event-render/EventRenderUtils.js
```diff
@@ -31,10 +43,18 @@ export function getVisibleEventEls() {
   return $(`.${EVENT_CLASS}:visible`)
 }
 
+export function getDayEls(){
```
arshaw: why is a utility related to day elements in the **events** file?

tests/automated/legacy/ListView.js
```diff
@@ -40,7 +53,7 @@ describe('ListView rendering', function() {
       it('filters events through eventRender', function() {
         var options = {}
         options.eventRender = function(arg) {
-          $(arg.el).find('.fc-event-dot').replaceWith('<span class="custom-icon" />')
+          replaceEventElDotElWithEl($(arg.el),'<span class="custom-icon" />')
```
arshaw: it's good to use a util function to query the event el and the dot el, but no need to make a util function that does the replaceWith.

tests/automated/legacy/ListView.js
```diff
@@ -430,35 +443,35 @@ describe('ListView rendering', function() {
     initCalendar(options)
     currentCalendar.changeView('listWeek')
 
-    expect($('.fc-list-item').length).toBe(1)
+    expect(getListItemElsCount()).toBe(1)
```
arshaw: the classname here is list-item, yes, but they actually represent events. so can you rename to `getEventCount()` and the like?

tests/automated/legacy/agenda-view.js
arshaw: maybe in future versions the element that used to be a `th` turns into a `div`. `getHeaderCellEls` would start returning a collection of `div`s instead of a collection of `th`s. the `getHeaderCellEls` is in charge of worrying about that type of element type it is. the caller shouldn't care.

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -1,38 +1,110 @@
...
+
+export function getMoreEl(){
+  return $(`.${MORE_CLASS}`)
+}
+
+export function getMorePopoverEl(){
+  return $(`.${MORE_POPOVER_CLASS}`)
+}
+
+export function getMorePopoverElTitleEl(){
```
arshaw: it would be better to have a util that returns the text of the popover el. just `getMorePopoverTitle`

jstacoder:
oh, just saw the comment about making a new PR, i wont add anymore new files to this one

tests/automated/legacy/ListView.js
jstacoder: but since the class is specific to the list view, (so would be different from the standard class for events) should it be more geared towards the list-view? ie: `getListEventElCount`?

tests/automated/event-render/EventRenderUtils.js
```diff
@@ -31,10 +46,34 @@ export function getVisibleEventEls() {
...
+
 export function getEventEls() {
   return $(`.${EVENT_CLASS}`)
 }
 
+export function getBackgroundEventEls() {
+  return $(`.${BACKGROUND_EVENT_CLASS}`)
+}
+
+export function getDayGridRowElAtIndexBackgroundEventEls(index) {
```
arshaw: what the heck do these functions do? names are crazy

tests/automated/event-render/EventRenderUtils.js
```diff
@@ -57,6 +96,19 @@ export function getSingleEl() {
...
+  return els
+}
+
 export function getEventElTimeEl(el) {
-  return el.find('.fc-time')
+  return el.find(`.${TIME_CLASS}`)
+}
+
+export function getEventElsWithCustomClass(customClass) {
+  return $(`.${customClass}`)
```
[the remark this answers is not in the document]
jstacoder: so in case's where a test is just passing a test specific class to jQuery to test that its in the resulting calendar element should i just let it be? my idea was to isolate those cases with this function? or just make a test specific function?

tests/automated/event-render/EventRenderUtils.js
jstacoder: `getDayGridRowElAtIndex` returns the day-grid row at the given index and `getDayGridRowElAtIndexBackgroundEventEls` returns the background event el's from a day-grid row at a given index, i wasnt sure what to call it at that point, ideas?

jstacoder:
so i can keep using the same functions ive been writing on the new pull request, could you make a branch and merge my v4 changes? otherwise its going to show the entire difference...

tests/automated/event-render/EventRenderUtils.js
arshaw: a preliminary comment:

- this `EventRenderUtils.js` file is supposed to have general event element utilities that aren't specific to any particular view/component, but rather are specific to *all* views. if so, why are we putting *day-grid* utils in here? can you make a file called `event-render/DayGridEventRenderUtils.js` instead?

i have two solutions to making these functions less crazy:

- stop having *AtIndex functions. instead, always return the whole jquery object with all the elements, and the caller can just use `.eq(i)` to get the one they want. if we follow this, that would get rid of `getDayGridRowElAtIndex` and turn it into `getDayGridRowEls`.
- make the functions more composable. like this...

```js
function getBackgroundEventEls(containerEl) { // modify it to accept an optional container el
    return $(`.${BACKGROUND_EVENT_CLASS}`, containerEl)
}

// then, instead of getDayGridRowElAtIndexBackgroundEventElAtIndex, we do...
getBackgroundEventEls(
    getDayGridRowEls().eq(i)
).eq(j)
```

tests/automated/event-render/EventRenderUtils.js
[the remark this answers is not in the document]
arshaw: yeah, just let it be. if the caller defined the className, and it's looking in the DOM for that className, we can allow that. just allow it to query for `$('.custom-class')` directly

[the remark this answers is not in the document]
arshaw: where is `getEventElsWithCustomClass` being used? i don't see it in this PR...

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -1,38 +1,138 @@
...
+
+export function getMoreEl() {
+  return $(`.${MORE_CLASS}`)
+}
+
+export function getMorePopoverEl() {
+  return $(`.${MORE_POPOVER_CLASS}`)
+}
+
+export function getMorePopoverElTitleEl() {
```
arshaw: i don't think you say my previous comment about this. it is...

> it would be better to have a util that returns the text of the popover el. just getMorePopoverTitle

tests/automated/lib/ViewUtils.js
```diff
@@ -0,0 +1,14 @@
+const VIEW_CONTAINER_CLASS = 'fc-view-container'
+
+
+export function getViewContainerEl() {
+  return $(`.${VIEW_CONTAINER_CLASS}`)
+}
+
+export function getViewContainerElHeight() {
```
arshaw: we might be taking this util function idea a little too far. i don't think we need `getViewContainerElHeight` and `getViewContainerElWidth`, the caller could just do `getViewContainerEl().height()` instead. our goal is not to abstract away the whole DOM, but rather extract away querying the DOM for selectors that are dependent upon HTML that fullcalendar generates.

tests/automated/lib/ListViewUtils.js
```diff
@@ -0,0 +1,60 @@
...
+
+export function getListEventElTitleEl(el) {
+  return el.find(`.${LIST_ITEM_TITLE_CLASS}`)
+}
+
+export function getListEventElTimeEl(el) {
+  return el.find(`.${LIST_ITEM_TIME_CLASS}`)
+}
+
+export function getListEventElTitleElText(el) {
```
arshaw: this can simply be called `getListEventElTitle` ("title" implies that text will be returned)

tests/automated/lib/ListViewUtils.js
```diff
@@ -0,0 +1,60 @@
...
+  return getListEventElTitleEl(el).text()
+}
+
+export function getListEventElTimeElText(el) {
```
arshaw: this can simply be called `getListEventElTimeText`

tests/automated/view-render/DayGridRenderUtils.js
```diff
@@ -102,7 +102,7 @@ export function getMorePopoverEl() {
   return $(`.${MORE_POPOVER_CLASS}`)
 }
 
-export function getMorePopoverElTitleEl() {
+export function getMorePopoverTitle() {
```
arshaw: this should return TEXT, not a DOM element. the naming convention is that if the method ends in "El", it will return an element, if not, it will return something else, in this case text.

once you make this change, places that call `getMorePopoverTitle` won't need to do `.toHaveText('asdfasdf')`. they can just do `.toBe('asdfasdf')`

once you make this change, we are good to merge.

PR #4233, opened 2018-07-03
Set moment and jquery and as peerDependencies

### LINKED ISSUE #4136: move jQuery and moment to peerDependencies
Hi,

Currently, both jQuery and moment are set as `dependencies`. With current package managers, it might end up in having different versions of these libraries installed without any warning.
Would it be possible to set both of them as `peerDependencies`? Popular libraries and framework are [doing it](https://github.com/angular/angular/blob/master/packages/core/package.json)!

[Blog post](https://nodejs.org/en/blog/npm/peer-dependencies/), and [documentation](https://docs.npmjs.com/files/package.json#peerdependencies) just in case.

alexander-schranz:
fixes #4136


acerix:
Thanks for the PR

alexander-schranz:
@acerix can't find out whats wrong with the ci :/

arshaw:

PR #4243, opened 2018-07-10
Fix type definition for eventAllow callback

### LINKED ISSUE #4236: TypeScript: Wrong parameter type for eventAllow
According to the TypeScript type definitions the `eventAllow` function should be called with a `DropInfo` and an `Event` parameter. This doesn't make sense, and indeed the function is actually called with a `DropInfo` and an `EventObjectInput` parameter.

Since type errors cannot be tracked in a CodePen I created a StackBlitz for this: https://stackblitz.com/edit/typescript-kjxyih . The compiled code does work, but if you wait for IntelliSense to show up, you'll see the type error in the editor.

Fixing this should be a one-liner. I'll gladly provide a PR if you like.

kremerd:
Fixes #4236.

### DIFF
--- a/src/types/input-types.ts
+++ b/src/types/input-types.ts
@@ -212,7 +212,7 @@ export interface OptionsInputBase {
   dragScroll?: boolean
   eventOverlap?: boolean | ((stillEvent: EventObjectInput, movingEvent: EventObjectInput) => boolean)
   eventConstraint?: ConstraintInput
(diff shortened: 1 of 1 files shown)


kremerd:
The automated checks are failing for reasons unrelated to this PR. Check #4245 for details.

arshaw:

PR #4245, opened 2018-07-10
Rewrite package-lock.json with Node 10.

kremerd:
Try a CI run with a freshly created package-lock.json file. [Search results](https://github.com/nodejs/node/issues/20285) suggest that might fix the pipeline issues that occur since the CI is using Node 10.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,13 +13,13 @@
     "@types/jquery": {
       "version": "2.0.47",
       "resolved": "https://registry.npmjs.org/@types/jquery/-/jquery-2.0.47.tgz",
-      "integrity": "sha1-lmWhV1Gd1I0ln5SsZw0zKlZWHAA=",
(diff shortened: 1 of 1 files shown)


acerix:
Thanks for the PR. Is there a github issue for the problem this fixes? I'm not sure what you mean by "pipeline issues".

kremerd:
No, there's no issue for this yet. I just wanted to figure out why none of the recent pull requests (most importantly #4243 🙄) passes the automated checks. Apparently the reason for this is that the CI was upgraded to Node 10 and somehow this messed up npm install. This PR shows one way to fix things up.

Originally I was just planning to mention this in #4243 to account for the failing checks. Should I raise an issue instead?

acerix:
Thanks for the info. Usually we want an issue for every PR but this one is not really a problem with the calendar so I don't think that is necessary.

PR #4329, opened 2018-10-03
Update ja.js

kusanaginoturugi:
I modified it so that the meaning passed.


arshaw:

PR #4578, opened 2019-03-25
feat: add typings to defaultView in OptionsInput

sheikalthaf:


arshaw:
i'm not sure this is the best idea because plugins are able to provide new view names.

PR #4589, opened 2019-03-28
adding viewSkeletonRender function to OptionsInputBase interface

vinicius0026:
As instructed on a fullcalendar/scheduler issue (https://github.com/fullcalendar/fullcalendar-scheduler/issues/508) I'm using `viewSkeletonRender` function to replace the Resource Header element after the initial render.

This breaks the TypeScript build, because `viewSkeletonRender` is not on the `OptionsInput` interface.

This PR just adds `viewSkeletonRender` to the `OptionsInputBase` interface.


arshaw:
fixed in [v4.0.2](https://github.com/fullcalendar/fullcalendar/releases/tag/v4.0.2)! thanks!

PR #4646, opened 2019-04-25
Easier dev setup by making Sass a devDependency

### LINKED ISSUE #4651: Make FullCalendar easier to setup and build by switching to JS Sass
Currently, you can't clone the fullcalendar repo, run npm install, and then build because a global install of Sass is required. 

Furthermore, you must use the Ruby version of Sass (`gem install sass`) instead of the JS version (`npm i -g sass`) because the `--nosourcemap` command-line parameter in package.json is `--no-source-map` in the JS variant of Sass CLI.  Having on a non-JS dependency complicates things (I assume especially for Windows where Ruby won't be installed by default). It's also possible that using this separate dependency might be the cause of #4626, although I'm not a Windows user so can't repro.

Also FYI: the Ruby version of Sass (the one used in the Travis config) is end-of-lifed, so even if we don't swtich to a JS Sass, using the Ruby one is probably something to move away from. Here's the output I just got from installing Ruby Sass on my Mac: 
```
Ruby Sass has reached end-of-life and should no longer be used.

* If you use Sass as a command-line tool, we recommend using Dart Sass, the new
  primary implementation: https://sass-lang.com/install

* If you use Sass as a plug-in for a Ruby web framework, we recommend using the
  sassc gem: https://github.com/sass/sassc-ruby#readme

* For more details, please refer to the Sass blog:
  https://sass-lang.com/blog/posts/7828841

```

Anyway, requiring a global, non-JS dependency seems to add unnecessary complexity to the setup, build, and dependency-updating process. It'd be easier if Saas was a local devDependency that could be installed and updated just like all other dev dependencies.

### LINKED ISSUE #4671: Could not find an option named "sourcemap".
### Bug Reports
To reproduce the issue, update your global sass version and try running ```npm run sass```.
I had this issue come up when installing a clean repo and installing the latest version of sass globally.

This is failing because we are still using the ```--sourcemap``` option and need to switch to using the ```--no-source-map``` option.

Please see [sass-lang's documentation](https://sass-lang.com/documentation/cli/dart-sass#no-source-map) for this option if you have any questions.

justingrant:
Currently, you can't clone the fullcalendar repo, run `npm install`, and then build because a global install of Sass is required. Furthermore, you must use the Ruby version which further complicates things for devs who don't use Ruby or who are on Windows. 

This global Ruby install seems like an unnecessary hurdle for new contributors or in new dev environments.  It'd be easier if Saas was a local devDependency.

Fixes #4651. Fixes #4671.

This PR: 
1. Adds Saas as a devDependency
2. Changes the "sass" script command line in package.json. `sourcemap=none` doesn't work with the JS Sass CLI, so changed it to `--no-source-map` which does work for the JS version.
3. Removes the global install of Sass from travis.yml


justingrant:
Hmm - looks like installing Sass locally might not be working with the Travis build, or there must be some other reason why a local dependency won't work. I'll close this PR if so and just install Sass globally for testing my PR.

acerix:
Thanks for the PR, but please also submit a feature request under Issues, then link this PR to it.

justingrant:
Hi @acerix - done and linked. Also pushed another commit to remove the (now unnecessary) Ruby install of Sass from the Travis config.  Let me know if you want me to rebase & squash these commits.

arshaw:
i actually implemented this when i converted the repo into a monorepo over the last month or so. in master. thanks for the work anyway

arshaw:
This was fixed in 4.3.0 though please start using 4.3.1
Releases: https://github.com/fullcalendar/fullcalendar/releases

PR #3694, opened 2017-05-31
Year View with 12 Mini-Calendars #1140

dusanm65:

Issue #1140, opened 2015-08-19
Year View with 12 Mini-Calendars


arshaw:
Originally reported on Google Code with ID 871

```
The attach file is a year view sample.
I made it by fullcalendar version.


```

Reported by `kebin.liu` on 2011-03-28 09:10:48

<hr>
- _Attachment: [fullCalendar15.rar](https://storage.googleapis.com/google-code-attachments/fullcalendar/issue-871/comment-0/fullCalendar15.rar)_

Imported with **51** stars.

piyushchauhan2011:
:+1:

phammer:
@arshaw : Any more or less official plan to release such a year view in the official release? We started to prototype with tanguy.pruvot's year view.

valteraki:
Hello!  I am using this and till now all went smooth ... But now i encountered a problem  .. is it possible to choose a event  by yearview (considering there will always only be 1 event for 1 day) and for example go to a page or get the title (url:) all this only on the yearview...

LorenGr:
Hi all, tanguy's addition to fullcalendar is absolutely great, fullcalendar however is moving forward in versions and year view is getting outdated. Please help me update tanguy's yearview with the latest fullcalendar version. I already did work to adapt to the updated fullcalendar grid and view api's. Please find it here : http://plnkr.co/edit/5U6vua
I am not managing to get the renderSelection mechanism working. And the css is missing, but most of yearview is now working with the latest fullcalendar version, check out the plunker above.

freshman21:
Hi, any luck on putting this to work on the newer version? I'm trying to adapt the plnkr example LorenGr posted but i keep getting an error on this line:

dayGrid.coordMap.containerEl = dayGrid.scrollerEl; // constrain clicks/etc to the dimensions of the scroller

it says coordMap is undefined, what am i missing? I'm not very experienced in this kind of stuff

G89Angel:
Would be great to have this in the new 3.0.1 version. Any plans of implementing it anytime soon?

espen:
@G89Angel 3.0.1 is already released. https://github.com/fullcalendar/fullcalendar/releases/tag/v3.0.1

arshaw:
A year view can be created via a **list view** but I believe this ticket is concerned with rendering a year as 12 tiled mini month-view calendars.

spookycookie27:
What would be the correct way to get the year view to start on the current month, rather than defaulting to January ?

meoz:
[...]

PR #4667, opened 2019-05-03
Created year-view.html demo

markclowes:
Year view demo with extra css and dayRender logic

### DIFF
--- a/demos/year-view.html
+++ b/demos/year-view.html
@@ -1,0 +1,220 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset='utf-8' />
+<link href='../packages/core/main.css' rel='stylesheet' />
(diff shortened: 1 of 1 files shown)


acerix:
Thanks for the PR, is the demo as described here?

https://github.com/fullcalendar/fullcalendar/issues/1140

markclowes:
It is kind of similar to #1140. It is a full year view in a single calendar window, not 12 mini calendars. Extra classes are added so each month can be styled differently.

CalvinWu4:
Can you post a screenshot?

arshaw:
this seems to be the demo for full-year view, but where is the actual implementation of full-year view?

arshaw:
from this PR i assume? https://github.com/fullcalendar/fullcalendar/pull/3694

Issue #678, opened 2015-08-19
Issue when tooltip is loaded using AJAX


arshaw:
Originally reported on Google Code with ID 408

```
I'm using full calendar with wTooltip plugin. I'm loading the contents of 
the tooltip via an AJAX call. I have also wired different functions on the 
click event of previous and next buttons. The previous and next clicks 
work fine untill the hover on an event and the tooltip is rendered. After 
that even if the click the buttons no event is fired. I'm attaching the 
code and a screenshot of my page. It is the same with the qtip plugin too.
I get an error saying 'Microsoft JScript runtime error: 'settings' is null 
or not an object' atattached error screenshot.


```

Reported by `hns.mailbox` on 2010-03-18 01:28:38

<hr>
- _Attachment: [script.rtf](https://storage.googleapis.com/google-code-attachments/fullcalendar/issue-408/comment-0/script.rtf)_
- _Attachment: [calendar.bmp](https://storage.googleapis.com/google-code-attachments/fullcalendar/issue-408/comment-0/calendar.bmp)_
- _Attachment: [error.bmp](https://storage.googleapis.com/google-code-attachments/fullcalendar/issue-408/comment-0/error.bmp)_

PR #4648, opened 2019-04-25
Go back to Node 11 until Travis has Node 12 support

justingrant:
All builds are failing because of a problem with Travis's support of Node 12 yet. https://discourse.elm-lang.org/t/travis-build-failing-and-other-occassions/3546

This PR switches the Travis build config from using the latest Node to instead stick to Node 11 so that builds will work again.

When Travis builds on Node 12 are working, this change should be reverted.

Fixes #4649

PR #4669, opened 2019-05-07
Updated sass script

wagnermaciel:
We are having the same problem as [ColorlibHQ Issue #678](https://github.com/ColorlibHQ/gentelella/issues/687).

```npm run sass``` is failing because we are still using the ```--sourcemap``` option.

This PR is a simple update to using ```--no-sourcemap``` instead of ```--sourcemap=none```

Please see [sass-lang's documentation](https://sass-lang.com/documentation/cli/dart-sass#no-source-map) for this option if you have any questions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "tsc-watch": "tsc --watch --sourceMap",
     "rollup": "rollup -c --environment BUILD:production",
     "rollup-watch": "sleep 5 && rollup -c --environment BUILD:development --watch",
-    "sass": "sass --update --sourcemap=none $(bin/sass-args.js)",
+    "sass": "sass --update --no-source-map $(bin/sass-args.js)",
     "sass-watch": "sass --watch $(bin/sass-args.js)",
     "test": "karma start karma.config.js --auto-watch",
     "test-single": "karma start karma.config.js --browsers ChromeHeadless_custom --single-run",


wagnermaciel:
For whoever reviews this, please note that your tests are failing due to the fact that [Gulp 3 is no longer supported and is broken on Node 12](https://github.com/gulpjs/gulp/issues/2324).

acerix:
Thanks for the PR.  Please submit a ticket under Issues with a [runnable, stripped-down demonstration](http://fullcalendar.io/wiki/Reduced-Test-Cases/) or the steps to reproduce the error, then link this PR to it.

acerix:
Sorry, I just saw that you did:

https://github.com/fullcalendar/fullcalendar/issues/4671

justingrant:
@wagnermaciel - I suspect this PR will break TravisCI builds because it doesn't update the global Sass install in Travis config. It'll be easy to see this after Travis builds are working again via #4648, but in the meantime it's easy to make the one-line Travis change in #4648 so that you can test your PR on a functional Travis config.

Also, would this PR require all contributors to upgrade Sass from the (old) Ruby gem version to a newer one?  If yes, would this be considered a breaking change?

BTW, would it make more sense to switch (via #4646) to making Sass a local devDependency which wouldn't require any global install changes for contributors, would also make it easier for new contributors, and would remove Saas versioning as a variable that could cause future problems?  Or is there a benefit to sticking with a global install?

arshaw:
i merged this PR but then ended up reverting the commit because it didn't work on my machine. I have the non-dart version of sass installed, which doesn't know about `--no-sourcemap`

solving this problem the "right way" is to implement #4651 (switch to a more portable version of SASS). I plan to do that for the next release

PR #4720, opened 2019-06-07
fix sourcemaps + publish /src to npm

### LINKED ISSUE #4719: Include working sourcemaps in dist
FullCalendar includes sourcemaps but there are multiple problems that prevent them from being useful, especially with file-based debuggers like VSCode: 
* the copyright/license banner added by gulp to the bundle file after sourcemaps are generated throws off line numbering
* the original TS source files are not included in FullCalendar npm packages. This means that file-based debuggers like VSCode won't be able to fully use debugging features, like setting breakpoints on FullCalendar files when the debugger is not running.  
* even if the files were installed, the `sources` array in the sourcemap points to paths that don't exist when FullCalendar is installed from npm, which will also break VSCode debugging. For example, below is the first few lines of the 4.2 sourcemap (https://unpkg.com/@fullcalendar/core@4.2.0/main.js.map), showing paths like `"../../src/core/util/dom-manip.ts"` that point to the build-time paths, not real paths under the npm install directory of any FullCalendar package.
```json
{  
   "version":3,
   "file":"main.js",
   "sources":[  
      "../../src/core/util/dom-manip.ts",
      "../../src/core/util/geom.ts",
      "../../src/core/util/scrollbars.ts",
      "../../src/core/util/dom-geom.ts",
      "../../src/core/util/dom-event.ts",
```

To repro: 
1) Using VSCode, open any app that installed FullCalendar from npm
2) Debug the app, setting a breakpoint in app code that calls into FullCalendar
3) Step into that call. You might have to Step Into several times to get past the module-loading shims. 

Expected: Debugger shows the first line of the method in the original TS source file
Actual: Debugger shows a line in a transpiled JS file. If you manually copy source files into the folders expected by the sourcemap, then the debugger will show the wrong line in a TS file.

I've built a PR that will address these problems. Will file it shortly. 

P.S. - to clarify, sourcemaps for debugging FullCalendar itself (for contributors) probably work just fine. The problem is when developers who use FullCalendar as a dependency want to debug into it.

justingrant:
Fixes #4719.  This PR improves the debugging story for app devs who use FullCalendar, so that devs (especially users of VSCode) can easily debug into the original FullCalendar source inside node_modules. 

This PR makes a few changes to the build process but doesn't make any runtime source code changes.  Here's what's new:
* Generates sourcemaps in all builds, not just dev builds.  This doesn't change the code that's downloaded in production bundles (especially not minified bundles) but simply makes sure that sourcemaps are always available.
* Updates rollup config to copy the original TS/Sass/etc source from `/src/packagename/` folders into each package's root folder. For example, `/src/core/**` gets copied into `/dist/core/src/**`
* Also copies locales source code (which is duplicated in every package) into a `/src/locales` folder of each package
* Fixes the paths in the `sources` array in sourcemaps so that sourcemaps point to the correct relative folder where the source was copied into in the steps above
* Removes newlines from the copyright/license banner that's prepended to bundle source files. This prevents line numbering from getting messed up in the sourcemap because adding the banner no longer changes the number of lines in the bundle js file.
* Ensures that license, readme, package.json, and .d.ts files are copied into the archive folders as part of the build process. This was needed for anyone who wants to build FC locally and npm install from that local build.  Previously they were missing. I'm actually unsure how publishing to npm ever worked given these missing files, unless the actual publishing to npm was using something different from `/bin/build-release.sh` that was checked into master. 

I've validated that the PR is working with the VSCode debugger as well as Chrome Dev Tools' debugger.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "native-promise-only": "^0.8.1",
     "rollup": "1.8.0",
     "rollup-plugin-commonjs": "^9.0.0",
+    "rollup-plugin-copy-glob": "^0.3.0",
     "rollup-plugin-multi-entry": "^2.1.0",
     "rollup-plugin-node-resolve": "4.0.1",
     "rollup-plugin-sourcemaps": "^0.4.2",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -7,14 +7,7 @@ import multiEntryArray from './rollup-plugin-multi-entry-array'
 import sourcemaps from 'rollup-plugin-sourcemaps'
 import rootPackageConfig from './package.json'
 import tsConfig from './tsconfig.json'
-
-let isDev
-if (!/^(development|production)$/.test(process.env.BUILD)) {
-  console.warn('BUILD environment not specified. Assuming \'development\'')
-  isDev = true
-} else {
-  isDev = process.env.BUILD === 'development'
-}
+import copy from 'rollup-plugin-copy-glob'
 
 let packageGlobals = {
   luxon: 'luxon',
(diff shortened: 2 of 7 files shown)


arshaw:
as of v5, the source files are encoded into the external sourcemaps. no `src` directory needed (i found this was much easier for consumption). thanks anyway for this PR @justingrant

PR #4731, opened 2019-06-17
Respected FirstDay setting even with week calc ISO

laoneo:
According to the [docs](https://fullcalendar.io/docs/firstDay) when weekNumberCalculation is set to 'ISO' it defaults to firstDay setting to Monday. But with the current code the firstDay setting is ignored when weekNumberCalculation is ISO.

### Testing instructions
- Load a calendar
- Set the firstDay setting to 3
- Set the weekNumberCalculation setting to 'ISO'

The result is that it shows Monday as first day in week.

Pen to test it is here https://codepen.io/anon/pen/PrNdwo

### DIFF
--- a/src/core/datelib/env.ts
+++ b/src/core/datelib/env.ts
@@ -63,12 +63,12 @@ export class DateEnv {
     this.weekDow = settings.locale.week.dow
     this.weekDoy = settings.locale.week.doy
 
-    if (settings.weekNumberCalculation === 'ISO') {
-      this.weekDow = 1
-      this.weekDoy = 4
-    }
     if (typeof settings.firstDay === 'number') {
       this.weekDow = settings.firstDay
+    }
+    else if (settings.weekNumberCalculation === 'ISO') {
+      this.weekDow = 1
+      this.weekDoy = 4
     }
 
     if (typeof settings.weekNumberCalculation === 'function') {
--- a/src/core/datelib/env.ts
+++ b/src/core/datelib/env.ts
@@ -66,7 +66,8 @@ export class DateEnv {
     if (settings.weekNumberCalculation === 'ISO') {
       this.weekDow = 1
       this.weekDoy = 4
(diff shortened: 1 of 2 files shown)


acerix:

laoneo:
Opened an issue #4734. Didn't find a chapter in the docs that I have to open an issue too. Did I miss that or is it not documented?

acerix:
I'll have check on that process since the docs doesn't say a separate issue is required:

https://github.com/fullcalendar/fullcalendar/blob/master/CONTRIBUTING.md#contributing-bugfixes

laoneo:
I don't want to put pressure on this, but can you guys give a rough estimate when this pr got reviewed and when ok, merged and shipped with an official release? It's a very disturbing bug and I need to know if I should fix it by my own for the time being or if it makes sense to wait.

arshaw:
hey @laoneo, we'll get this released in the next minor.

can you write a simple test? right around here:
https://github.com/fullcalendar/fullcalendar/blob/master/tests/automated/legacy/firstDay.js#L45

Also, can you change the ordering if the if statements and use an else-if? it should be:

```js
    if (typeof settings.firstDay === 'number') {
      this.weekDow = settings.firstDay
    }
    else if (settings.weekNumberCalculation === 'ISO') {
      this.weekDow = 1
      this.weekDoy = 4
    }
```

laoneo:
Honestly I try to avoid else statements, because of simplicity (some even go that far to treat it as a code smell) . Is it really needed?
I can make the test tomorrow.

arshaw:
It will work just the way you have it, but an `if else` statement is better because it prevents unnecessary code execution and also expresses to the reader that only one of treatments to `weekDow`/`weekDoy` should happen. I don't share the same negative opinions about else/ifelse statements that you do.

laoneo:
Changed order

laoneo:
What did I do wrong that travis failed? When I apply the patch locally then the calendar works as expected.

PR #4781, opened 2019-07-18
Should fix ssr problem "element is not defined" with nuxtjs and nextjs

Bornegrim:
Should fix ssr problem "element is not defined" with nuxtjs and nextjs

### DIFF
--- a/src/core/util/dom-manip.ts
+++ b/src/core/util/dom-manip.ts
@@ -118,7 +118,7 @@ export function removeElement(el: HTMLElement) {
 
 // from https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
 const matchesMethod=function(s) {
-  var matches = (this.document || this.ownerDocument).querySelectorAll(s),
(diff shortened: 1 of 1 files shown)


arshaw:
this change will cause all browsers to always use the really slow fallback provided for the "matches" method.

On a more fundamental level, why doesn't nextjs/nuxtjs support `Element.prototype.matches` nor `Element.prototype.closest` ?

haotangio:
I upgraded to nextjs 9.0.2 to be able to build. Just a temporary workaround.

brendanmckenzie:
We are experiencing the same issue using ReactJS.NET.  

`Element` does not exist in the global namespace, just like `document` and `window`.

zorrme:
Sam with angular universal, we should sanitize use of `document`, `window` and `Element`

arshaw:
I'm questioning whether it's even a good idea to allow FullCalendar to render in an SSR environment: https://github.com/fullcalendar/fullcalendar/issues/4784#issuecomment-652559090

I'm closing this PR until we come to a consensus with that.

Issue #4957, opened 2019-08-02
npm run build fails on fresh install


Invincibear:
### Bug Report

On a CentOS7 server, I performed the following steps [per instructions here](https://github.com/fullcalendar/fullcalendar/blob/master/CONTRIBUTING.md):
```
cd ~/httpdocs
git clone git://github.com/fullcalendar/fullcalendar.git
cd fullcalendar
sudo npm install -g sass yarn gulp
npm install
```
`npm run build` & `npm run watch` output the following errors:

[redacted successful output up until this point]
> [rollup] created packages/core/dist/locales-all.js in 71ms
> [rollup]
> [rollup] tmp/tsc-output/packages?(-premium)/__tests__/src/globals.js, tmp/tsc-output/packages?(-premium)/__tests__/src/**/*.js → tmp/tests.js...
> [rollup] **[!] Error: 'config' is not exported by ../../node_modules/@fullcalendar/core/main.js**
> [rollup] https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
> [rollup] tmp/tsc-output/packages/__tests__/src/hacks.js (1:9)
> [rollup] 1: import { config } from '@fullcalendar/core';
> [rollup]             ^
> [rollup] 2: beforeEach(function () {
> [rollup] 3:     // On real devices, when a click-like touch interaction happens, there is a preiod of time where mouse events
> [rollup] Error: 'config' is not exported by ../../node_modules/@fullcalendar/core/main.js
> [rollup]     at error (/var/www/vhosts/warmapcalendar.com/httpdocs/fullcalendar/node_modules/rollup/dist/rollup.js:3598:30)
> [rollup]     at Module.error (/var/www/vhosts/warmapcalendar.com/httpdocs/fullcalendar/node_modules/rollup/dist/rollup.js:14470:9)
> [rollup]     at handleMissingExport (/var/www/vhosts/warmapcalendar.com/httpdocs/fullcalendar/node_modules/rollup/dist/rollup.js:14397:21)
> [rollup]     at Module.traceVariable (/var/www/vhosts/warmapcalendar.com/httpdocs/fullcalendar/node_modules/rollup/dist/rollup.js:14739:17)
> [rollup]     at ModuleScope.findVariable (/var/www/vhosts/warmapcalendar.com/httpdocs/fullcalendar/node_modules/rollup/dist/rollup.js:13408:37)
[...]

PR #4966, opened 2019-08-09
Note about compiling FC

Invincibear:
After spending about a whole day of troubleshooting why FC wouldn't compile, I eventually figured out that if any parent directories contained dependencies for @fullcalendar/* libraries NPM would fail to compile. This warning hopefully saves somebody else from hours/days of frustration & troubleshooting

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,6 +64,7 @@ If you want to clean up the generated files, run:
 
 	npm run clean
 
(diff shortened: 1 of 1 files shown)


arshaw:
@Invincibear I'm sorry about that, must have been really annoying.

Is this the root cause of your problems in #4957?

There's no way for us to prevent node module resolution from looking in node_modules directories in the parent, however, if `npm install` was correctly ran (which runs `monorepo install` which wires up the package's local references to the `@fullcalendar/*` packages), then these parent directories would not be searched.

Did you get any errors on `npm install` ?

PR #4974, opened 2019-08-13
made GOOGLE_API_BASE configurable, fixes #4953

### LINKED ISSUE #4953: Make API_BASE in Google Calendar configurable
### Feature Requests
I use a proxy for Google Calendar API and need to change the API_BASE const to this proxy-url.

It would be great if API_BASE could be added to the props and optionally be set from outside like the API key.

https://github.com/fullcalendar/fullcalendar/blob/master/packages/google-calendar/src/main.ts :
```
// TODO: expose somehow
const API_BASE = 'https://www.googleapis.com/calendar/v3/calendars'
```
There is also a TODO there, so i guess this behaviour was desired before.

ivoba:
This makes the GOOGLE_API_BASE configurable in @fullcalendar/google-calendar and fixes https://github.com/fullcalendar/fullcalendar/issues/4953

Im not very experienced in typescript and monorepos, but i just gave it a shot.
Any advice is welcome.

### DIFF
--- a/packages/google-calendar/src/main.ts
+++ b/packages/google-calendar/src/main.ts
@@ -7,6 +7,7 @@ const STANDARD_PROPS = { // for event source parsing
   url: String,
   googleCalendarApiKey: String, // TODO: rename with no prefix?
   googleCalendarId: String,
+  googleCalendarApiBase: String,
   data: null
 }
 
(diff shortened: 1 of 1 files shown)


arshaw:
looks good

arshaw:
this is now in v4:

i've made a note to myself to add this to the docs

PR #5010, opened 2019-08-29
minor translation fix

AlexandraNY3:
A small translation fix for the greek (el) locale.


acerix:
Thanks for the PR.  Google translates both to "There are no events to show" so could breifly explain why your translation is correct?

arshaw:
this is released in v4.4.0:

Issue #4549, opened 2019-03-10
recurring event exceptions


nbcnc:
https://fullcalendar.io/docs/v4/recurring-events

Would love an ability to provide exceptions to the recurring event.  In other words, My Event occurs weekly between Friday March 8th and Friday March 29th **except** for March 15th.  Exceptions could be an array that lets us provide multiple dates.

```
{
daysOfWeek:  ['5'],
startTime: '12:00',
endTime: '13:00',
startRecur: '2019-03-08',
endRecur: '2019-03-30',
exceptions: ['2019-03-15']
}
```

flamber:
Hi @nbcnc 

Exclusions are supported in the `rrule` plugin: https://fullcalendar.io/docs/v4/rrule-plugin
```
{
  title: 'rrule string with exclusion',
  rrule: 'DTSTART:20190308T120000Z\nRRULE:FREQ=WEEKLY;UNTIL=20190330\nEXDATE:20190315T120000Z',
  duration: '01:00'
}
```
https://codepen.io/anon/pen/qvPxMZ?editors=0010

But it's only possible using the string-style - not the object-style - since the FullCalendar rrule plugin doesn't read the `EXDATE` or `EXRULE` parameters. (At least as of 4.0.0-beta.4)
Read more about the parameters (and other advanced stuff) on: https://github.com/jakubroztocil/rrule

arshaw:
thanks. we should add support for exdate/exrule as properties of the event object somehow

nbcnc:
Thanks @arshaw and @flamber.  I didn't notice that EXDATE was an option.  Adding it to the object style notation would definitely be helpful.  Here's an example of using EXDATE rule as a string in case anyone needs an example.

DTSTART:20190312T010000\nEXDATE:20190323T010000\nEXDATE:20190322T010000\nRRULE:FREQ=WEEKLY;INTERVAL=1;BYDAY=MO,TU,WE,TH,FR,SA,SU;UNTIL=20190330

Recurring starts on **2019-03-12 1AM** (MO-SU) until **2019-03-30** except **2019-03-22** and **2019-03-23**

Blackbaud-SeanStephens:
While looking into this topic, I noticed that EXRULE is considered deprecated by the RFC standard, so support for it in future versions of RRule may be dropped. The standard prefers the use of EXDATE only.

I only mention in case it influences the implementation decision.

Looking forward to this one!

ganySA:
[...]

Issue #4439, opened 2019-01-04
Implement RRuleSet's for more complex recurrence setups like excluding dates


piernik:
It's a feature request.
You wrote that You're using this lib form rrules: https://github.com/jakubroztocil/rrule
Could You add also rruleSets?

Why I need it? Let's assume that I have everyday tasks for a week. I forgot to do it on Monday but did it on Tuesday and have to do it on Wednesday and so on. 
If I use Rrule only, fullcalendar will mark all days with events. But on Tuesday shouldn't be event - it's done.

Will You consider my request ?:)

acerix:
Seems like it would be useful, especially since it's part of that lib.

https://github.com/jakubroztocil/rrule#rruleset-constructor

teriblus:
I need it too, I'll keep generating recurring on the server side until rruleset is available in fullcalendar.

arshaw:
btw, you can do this currently with an rrule string, like is mentioned in this comment: https://github.com/fullcalendar/fullcalendar/issues/4549#issuecomment-472228339

arshaw:
This has been implemented in v5.5.0. `exdate` and `exrule` are now top-level properties of the event object.

Please look at the updated docs: https://fullcalendar.io/docs/rrule-plugin

PR #5097, opened 2019-10-21
adds exdates, exdate, exrules and exrule as options for rrule with rr…

BoLaMN:
adds exdates, exdate, exrules and exrule as options for rrule with rruleset

implements #4439, #4549

### DIFF
--- a/packages/rrule/src/main.ts
+++ b/packages/rrule/src/main.ts
@@ -124,7 +124,7 @@ function parseRRule(input, dateEnv: DateEnv) {
 }
 
 function getExArray(input, key) {
-  let arr;
+  let arr
 
   if (input[key + 's']) {
     arr = input[key + 's']
--- a/packages/rrule/src/main.ts
+++ b/packages/rrule/src/main.ts
@@ -103,7 +103,7 @@ function parseRRule(input, dateEnv: DateEnv) {
       const rruleSet = new RRuleSet()
       rruleSet.rrule(rrule)
 
-      exrules.forEach((obj) => {
+      exrules.forEach(function (obj) {
         let rule = parseRRule(obj, dateEnv)
         if (rule && rule.rrule) {
           rruleSet.exrule(rule.rrule)
@@ -113,6 +113,7 @@ function parseRRule(input, dateEnv: DateEnv) {
       exdates.forEach(rruleSet.exdate)
 
       return { rrule: rruleSet, allDayGuess }
+    }
   }
 
   if (rrule) {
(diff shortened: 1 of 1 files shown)


BoLaMN:
currently failing on 

```
● handles multiple prop changes, include event reset
    expect(received).toBe(expected) // Object.is equality
    Expected: 2
    Received: 1
      79 |     expect(isWeekendsRendered(wrapper)).toBe(false)
      80 |     expect(eventRenderCnt).toBe(3) // +2
    > 81 |     expect(viewSkeletonRenderCnt).toBe(2) // +1
         |                                   ^
      82 |   })
      83 | })
      84 | 
      at toBe (FullCalendar.test.js:81:35)
``` 

but not why my changes would be effecting it at all

stunaz:
Any update on this? much appreciated

BoLaMN:
yeah not sure either, been over a month now without any review. 

Im guessing @arshaw is a one man band with pull requests and has priorities else where.

21Junipers:
@arshaw makes it clear that PRs might take a while to implement for various (completely) reasons. I hope it does not offend to state here how much rruleset support would bring to the table. Wish I could be useful in some way to get this in sooner rather than later. NOT complaining, this lib has come so far with v4 (Thanks!).

packages/rrule/src/main.ts
```diff
@@ -104,6 +123,20 @@ function parseRRule(input, dateEnv: DateEnv) {
   return null
 }
 
+function getExArray(input, key) {
+  let arr
+
+  if (input[key + 's']) {
```
arshaw: could you help me understand what's going on here? some comments would great.

BoLaMN: allows it to be defined as

exdate: new Date()
exdates: [ new Date(), new Date() ]
exrule: new Date()
exrules: []

in the options and transform it into an array on the return

21Junipers:
Dead?

BoLaMN:
It’s waiting for tests to be written, unfortunately my free time is negative 7 hours a day so till I can get around to it or someone else wants to contribute tests it’ll probably sit here for awhile

yalanyali:
This looks pretty good, I'd love to see it being published as well.

gongAll:
Recently I upgraded from v4 to v5 and some features broke. Basically passing rrules as string would allow for exdates to be passed on v4 and won't be allowed on v5 anymore.

Therefore, I see this pr as really important, since I cannot pass my project with v5 to production without it, which is a bummer. Would be glad to make tests for it, having some guidelines to follow.

arshaw:
Superseded by https://github.com/fullcalendar/fullcalendar/pull/5097, which has been released in v5.5.0. Please see my comment: https://github.com/fullcalendar/fullcalendar/issues/4439#issuecomment-748536028

Issue #5001, opened 2019-08-26
validRange and problem with drag&drop


szajdel:
To recreate problem please use link below:
https://codepen.io/SlawomirZajdel/pen/RwbpQVM?editors=0010#anon-signup

There is editable property set to 'true' on calendar options and I want to allow dragging only in selected range using validRange:
 { start: '2019-08-15',
 end: '2019-08-18' }
 There is no option to drag event with this setting.

When attaching validRange dynamically:
https://codepen.io/SlawomirZajdel/pen/JjPWLop?&editable=true&editors=001
there are errors in console log showing propably cause of issue:
Uncaught TypeError: Cannot read property 'isWithinClipping' of undefined
    at e.queryHitForOffset (main.min.js:6)
    at e.handleMove (main.min.js:6)
    at t.handleDragMove (main.min.js:6)
    at de (main.min.js:6)
    at t.triggerWith (main.min.js:6)
    at t.trigger (main.min.js:6)
    at t.r.onPointerMove (main.min.js:6)
    at de (main.min.js:6)
    at t.triggerWith (main.min.js:6)
    at t.trigger (main.min.js:6)

szajdel:
Yes, did't know that same id is for repeating event, thx.

Second example fixed now. There is event at 26-08-2019 and during drag start - there is validRange added from eventDragStart. When you try to move event to other place (like 2h before) there are console errors.

acerix:
I can see there is an error when changing `validRange` from `eventDragStart` so that seems like a bug.  To restrict where the event can be dragged you should use `eventAllow` though.

https://fullcalendar.io/docs/eventAllow

szajdel:
Any chance it will be fixed ?

microJ:
the same error: `TypeError: Cannot read property 'isWithinClipping' of undefined`

ziaulrehman40:
@szajdel Thankyou so much for the inspection and everything.
Our use case was in fullcalender's react package and I had something like:

```
eventConstraint={{
  start: Moment().format('YYYY-MM-DD'),
  end: '2200-01-01' // hard coded goodness unfortunately
}}
```

After reading through this issue I used a constant instead of inline declaring object, and it is now working without any errors.
So yeah, i think issue is with when eventConstraint changes due to some reason.
Pasted my use case so could be useful for anyone looking for same issue.

xjwalker:
`TypeError: Cannot read property 'isWithinClipping' of undefined`

[...]

PR #5145, opened 2019-11-22
fix: issue 5001

microJ:
fix error: `Cannot read property 'isWithinClipping' of undefined`
issue: https://github.com/fullcalendar/fullcalendar/issues/5001

### DIFF
--- a/packages/interaction/src/interactions/HitDragging.ts
+++ b/packages/interaction/src/interactions/HitDragging.ts
@@ -164,7 +164,7 @@ export default class HitDragging {
       let component = droppableStore[id].component
       let offsetTracker = offsetTrackers[id]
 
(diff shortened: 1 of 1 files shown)


arshaw:
thanks so much for this fix, but instead of introducing this hack to prevent a JS error, I want to see how v5 handles this. I'm fairly certain it will be fixed due to the way it has been rearchitected

jpehman:
Hi @microJ and @arshaw 
This error prevents the dragging of events in v4.4. Is there any chance you would be willing to merge in the fix, so that we can allow our users to drag events? This is guaranteed to generate support requests for our product. In our case it's also causing a weird shrinking of the fullcalendar. I would make the change locally, but fullcalendar is included via npm in our build process.

PR #5167, opened 2019-11-29
#4317 replace all style= with a data-attribute for CSP support

### LINKED ISSUE #4317: Does not support Content Security Policy (CSP) nonce, only unsafe-inline css
Cannot use "script-src 'nonce-...." only supports "style-src 'self' 'unsafe-inline';"

In fullcalendar.js starting at line 11408:
```
// Generates an HTML attribute string for setting the width of the axis, if it is known
    AgendaView.prototype.axisStyleAttr = function () {
        if (this.axisWidth != null) {
            return 'style="width:' + this.axisWidth + 'px"'; <--------- violation will not load
        }
        return '';
    };
```
calnoncecsponlyunsafeinline_15sep18

vandernorth:
Replaced all style-tags with a data-attribute that sets the style later. this is for CSP support and fixes #4317

I've tested this against the color-options on events. On a page with the `Content-Security-Policy` header set to `default-src 'self' data:;` you used to see CSP errors in the console and the color options didn't work. With this fix the CSP errors are gone and you see the selected color.

### DIFF
--- a/packages/core/src/Calendar.ts
+++ b/packages/core/src/Calendar.ts
@@ -167,11 +167,26 @@ export default class Calendar {
       this.renderableEventStore = createEmptyEventStore()
       this.bindHandlers()
       this.executeRender()
+      this.dataToStyle()
     } else {
       this.requestRerender()
     }
   }
 
+  dataToStyle() {
+    document.querySelectorAll('*[data-calendar-style]').forEach(node => {
+      const hasStyleAttribute = !!node.attributes['style']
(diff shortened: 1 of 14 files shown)


vandernorth:
I'm working on the tests

arshaw:
this PR has fallen out of date because the codebase has changed quite a bit (my fault), but I feel that we can close this because I have enough information to develop a fix based on #5556 and #4317

PR #5185, opened 2019-12-15
Azerbaijan Language Pack Added

rommelmamedov:
Is there any update on this? 🙂


arshaw:
this is released in v4.4.0:

HOWEVER, the browser must support the `az` locale in order for this to work. On my computer Firefox does, but Chrome doesn't for example. If you want more consistent support, it's advised to use the moment plugin.

PR #5236, opened 2020-01-29
French translation for "list" is "my schedule" but should be just "schedule"

Oviglo:
Just Planning not "Mon Planning" because it can be an other way (like other user)


arshaw:
this is released in v4.4.0:

PR #5293, opened 2020-03-05
Supress errors when interacting with calendar during rerendering

FreexD:
When interacting with calendar:
- dragging events, 
- selecting dates, 
- resizing events

during calendar rerendering due to:
- event source refetch happening,
- calendar initializing,
- rerender method being called programmatically

We can get different kinds of errors:
- `e.queryHitForOffset: Cannot read property 'isWithinClipping' of undefined`,
- `windowResize: undefined is not an object (evaluating 'e.target')`
- `t.e._splitIndividualUi : Cannot read property 'length' of undefined`,
- `Wt: Cannot read property 'fcSeg' of null`.

This is probably caused by trying to mutate events that are being removed from eventStore or are already removed. This PR only makes errors not appear in console. It doesn't fix inconsistencies that can be created during such interaction. 

For example:
- we're dragging an event,
- a refetch and rerender happens,
- the event that we're dragging gets deleted from the eventStore,
- new event (representing the deleted one) takes it's place in the eventStore,
- dragging placeholder still appears and indicates we're dragging,
- (thanks to the fix we're no longer getting errors in the console),
- we drop the dragged event, 
- eventDrop get's called and placeholder gets replaced with normal event,
- we end up with 2 events (representing the same one) appearing on the calendar at the same time.

I will provide a proper issue, repro and tests to my fix ASAP.

### DIFF
--- a/packages/core/src/Calendar.ts
+++ b/packages/core/src/Calendar.ts
@@ -882,7 +882,7 @@ export default class Calendar {
     if (
       !this.isHandlingWindowResize &&
       this.component && // why?
-      (ev as any).target === window // not a jqui resize event
+      (ev && (ev as any).target === window) // not a jqui resize event
     ) {
       this.isHandlingWindowResize = true
       this.updateSize()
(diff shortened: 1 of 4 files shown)


acerix:

EdmodoYunjiang:
Drag the same event multiple times in a short time, this event will be rendered twice

image

arshaw:
can't recreate this issue, so closing

Issue #4955, opened 2019-08-01
Date objects in rrule options are not converted to local time


martinsiden:
Hi there

I'm running Fullcalendar with the default `local` as timezone and I have RRule installed. I create and store RRule strings in my database which I pull back to Fullcalendar to render. Dates are stored as UTC in the database. 

When rendering an event using the RRule string like this:

```
{ 
      id: '2',
      title: 'Recurring', 
      rrule: 'DTSTART:20190801T080000Z\nRRULE:FREQ=WEEKLY;INTERVAL=1;UNTIL=20190810;BYDAY=MO,TU,WE,TH,FR,SA,SU',
      duration: '02:00'
    }
```

The `DTSTART` value isn't converted into local time like it is when you define it as an object like this:

```
{ 
      id: '2',
      title: 'Recurring', 
      rrule: {
        freq: 'weekly',
        interval: 1,
        byweekday: ['mo', 'tu', 'we', 'th', 'fr', 'sa', 'su'],
        dtstart: '2019-08-01T08:00:00Z',
        until: '2019-08-10'
      },
      duration: '02:00'
    }
```

If you use RRules `rrulestr` function to map the string into an RRule object, you explicitly have to call `toISOString()` on `dtstart` to make it work. 

```
{
 freq: rrule.options.freq,
 interval: rrule.options.interval,
 byweekday: ['mo', 'tu', 'we', 'th', 'fr', 'sa', 'su'],
 dtstart: rrule.options.dtstart.toISOString(),
 until: rrule.options.until.toISOString()
}
```

I guess this is expected behaviour due to the datetime format that's in the RRule string? Is there a better way of dealing with this issue? Can Fullcalendar take care of this for me?

Here's a sample repo (Angular app) for this:
https://github.com/martinsiden/fullcalendar-rrule

nigamabhay671:
It is a problem of Moment

acerix:
Thanks for the report, I made a demo to test it:

https://codepen.io/acerix/pen/OKxNyG?editors=0010

It seems like a bug to me; when `dtstart` in the rrule is a date object instead of a string it is not getting converted to local time.

paco3346:
Good news, this only took 6 hours to figure out :P

Anything that is a recurring definition is exempt for the built-in timezone adjustment. As seen here:
[...]

PR #5326, opened 2020-03-31
Make sure @fullcalendar/rrule parses rrules given as strings the same as rrules given as objects

joelzanden:
This fixes an issue with the rrule plugin parsing rrule strings without taking the current timezone into account.

This fix is suggested by @paco3346 in issue #4955 and I take no credit for it. It seems to fix a problem that I really want to see fixed, so I took the liberty of creating this pull request. The first one I've ever created. :)

### DIFF
--- a/packages/__tests__/src/datelib/rrule.js
+++ b/packages/__tests__/src/datelib/rrule.js
@@ -255,6 +255,27 @@ describe('rrule plugin', function() {
     expect(events[0].allDay).toBe(false)
   })
 
+  // https://github.com/fullcalendar/fullcalendar/issues/4955
+  it('can generate local dates when given an rrule string including EXDATE', function() {
+    const localDateToUTCIsoString = parseLocalDate('2018-09-04T05:00:00').toISOString()
+    const modified = localDateToUTCIsoString.replace('.000', '').replace(/[\-\:]/g, '')
+
+    const localExdate = parseLocalDate('2018-09-05T05:00:00').toISOString().replace('.000', '').replace(/[\-\:]/g, '')
+
+    initCalendar({
+      timeZone: 'local',
+      events: [
+        {
+          rrule: `DTSTART:${modified}\nRRULE:FREQ=WEEKLY\nEXDATE:${localExdate}`,
+        }
+      ]
+    })
+    let events = getSortedEvents()
+    expect(events.length).toBe(5)
+    expect(events[0].start).toEqualLocalDate('2018-09-04T05:00:00')
+    expect(events[0].end).toBe(null)
+    expect(events[0].allDay).toBe(false)
+  })
 
   function getSortedEvents() {
     let events = currentCalendar.getEvents()
--- a/packages/__tests__/src/datelib/rrule.js
+++ b/packages/__tests__/src/datelib/rrule.js
@@ -64,7 +64,7 @@ describe('rrule plugin', function() {
     expect(events[4].start).toEqualDate('2018-10-02T13:00:00Z')
   })
 
-  it('can expand monthly recurrence', function() {
+  it('can expand monthly recurrence when given an rrule object', function() {
     initCalendar({
       defaultView: 'dayGridMonth',
       now: '2018-12-25T12:00:00',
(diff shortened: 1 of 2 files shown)


joelzanden:
The new test should now fail.

It turns out that the function parseRRule() in the fullcalendar rrule plugin parses rrules in string format differently than rrules in object format, as pointed out by @martinsiden in issue #4955. I would dare call this behaviour a bug.

joelzanden:
I think this last commit is a valid fix.

joelzanden:
I have now added two tests to fullcalendar/rrule:
1. 'can generate local dates when given an rrule string'
2. 'can generate local dates when given an rrule string including EXDATE'

When an rrule is given in object format, the fullcalendar/rrule plugin does its fullcalendar magic and turns the dates into localized markers. But when an rrule is given in string format, the fullcalendar/rrule sends it through the `rrulestr()` method from the jakubroztocil/rrule dependency:

https://github.com/fullcalendar/fullcalendar/blob/0bb19a6c113bc3529ac7b6ee4b840f18e6ffaad5/packages/rrule/src/main.ts#L62-L63

The problem seems to be with the way jakubroztocil/rrule handles timezones:
https://github.com/jakubroztocil/rrule/blob/55be5f6523706e07abc6b97abc23dac3f47030a9/README.md#L169

The dates are returned in UTC (ending with a Z), but should be interpreted as dates in the local timezone. So I've added the boolean 'fromString' to the RRule object (orRRuleSet object)…
https://github.com/joelzanden/fullcalendar/blob/6b525ec51aeafc9e555fa211aff8e7bd80134661/packages/rrule/src/main.ts#L66-L68

…to be able to only apply the fix suggested by @paco3346 in issue #4955 :
https://github.com/joelzanden/fullcalendar/blob/6b525ec51aeafc9e555fa211aff8e7bd80134661/packages/rrule/src/main.ts#L42-L54

devchris:
Can this please be merged? 😁

DerStoffel:
Hi guys, 

I'd love this to be merged (especially for the angular version I really need this :) ).

Also, is there something I can do, to help you on that part?

tk250:
For me it would also help a lot if this gets merged

antoinerl:
Hi, it seems that it's not a new feature but a bug fix... and it passes all the checks. So could you please merge!
It's quite annoying for us to loose two hours each time we display the calendar!

Thank you for your work!

pratishshr:
Can this get merged, please? Thank you!

arshaw:
after looking at this PR further I noticed that [the new algorithm didn't process whole-day local dates correctly](https://github.com/fullcalendar/fullcalendar/commit/9b5a9adf27608d52b7683b03c3e5a570dbccd76d) so I [re-implemented the functionality by preprocessing the rrule string](https://github.com/fullcalendar/fullcalendar/commit/b40ed5a52b5e45f301ea4569503eafc69dcdd168)

arshaw:
this has been released in v5.2.0. please try it!

bodywizard:
I'm still getting issues with RRules that use RDate instead of dtstart/interval.  Example "RDATE:20201019T150000Z,20201020T150000Z".

bodywizard:
FYI I solved the problem I had by updating the "preparseRRuleStr" function in rrule/main.js

The original function preparsed the rrulestr only if it included dtstart, until, and exdate.  

This updated function will find any datestr in the entire rrule and preparse it.

`function preparseRRuleStr(str, dateEnv) {
    var isTimeUnspecified = null;
    function replaceRRuleDate () {
        let args = [...arguments], separator = args[2], str = args[3];
        var res = dateEnv.parse(str);
        if (res) {
            if (res.isTimeUnspecified) {
                isTimeUnspecified = true;
            }
            return separator + formatRRuleDate(res.marker);
        }
        else {
            return args[1];
        }
    }
    str = str.replace(/(([:,])(2[^,\n]*))/g, replaceRRuleDate);
    return { outStr: str, isTimeUnspecified: isTimeUnspecified };
}
`

Blindmikey:
Did this disappear in 5.4.0? 
In 5.3.2 this works fine: "rrule":"DTSTART:20210102T100000;RRULE:FREQ=DAILY;INTERVAL=1;COUNT=2"
In 5.4.0 onward it breaks the calendar with "Uncaught TypeError: Cannot read property 'isTimeUnspecified' of null"

arshaw:
@Blindmikey , please post a reduced test case elsewhere https://fullcalendar.io/reporting-bugs

PR #5334, opened 2020-04-03
updated path for locales

KuboF:


KuboF:
Hi @arshaw! Can I kindly ask you to process this PR? It is very straightforward, non-code change, so I think it could be simple.

I have already prepared also an Esperanto locale. But I currently can't make PR for it, because I have missed branches at my fork...

Issue #5348, opened 2020-04-11
v3 breaks with jQuery 3.5 due to security fix for invalid HTML tags


neroux:
I do realise version 3 is most likely not supported any longer, however I still reference to this issue, jquery/jquery#4664, as everyone upgrading such a setup to jQuery 3.5 will run into it due to a secuity fix they implemented.

https://github.com/jquery/jquery/issues/4664#issuecomment-612254616 essentially summarises it. FullCalendar is creating self-closing tags for tags which do no officially allow for that.

If it were possible to have a quick version 3.10.1 with that fix, I guess that would help everyone who hasn't rewritten their code for version 4 yet :)

neroux:
An example would be
https://github.com/fullcalendar/fullcalendar/blob/bad3b01082691191c8b7ce2158a454d6dc40360d/src/util.ts#L268

acerix:
Thanks, confirmed.

https://codepen.io/acerix/pen/VwvLRGK

Softsmiths:
Is it possible to fix version v2 version too?

arshaw:
this has been fixed and released!
https://github.com/fullcalendar/fullcalendar/releases/tag/v3.10.2
https://github.com/fullcalendar/fullcalendar-scheduler/releases/tag/v1.10.1

sorry @Softsmiths, we won't fix v2, only v3

Softsmiths:
@arshaw I think I fixed it myself in source code of 2.9.1 by changing closing html tags. It looks like it will work, at least for my use. Will do some testing.

PR #5361, opened 2020-04-16
fix: transform invalid self closing tags

lmancel:
This pull request is an attempt to solve the issue detailed in #5348 by transforming the self closing tags to "properly" closed tags whenever necessary.

Sorry in advance if this is not the proper way of contributing to this repository, we needed this fix for one of our projects so I did this in a hurry. Tell me if anything needs to be changed/improved.

There are 12 tests that are not passing (all located in events-gcal), but the same are also not passing when running the tests directly on v3.

If this suits you (and if it is needed) I could do the same modifications on v4.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3175,7 +3175,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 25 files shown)


aqjw:
Any updates?

arshaw:
this was addressed in 3.10.2: https://github.com/fullcalendar/fullcalendar/blob/v3/CHANGELOG.md#v3102-2020-05-13

thanks anyway for the PR @aqjw

Issue #5287, opened 2020-03-02
validRange input type does not allow callback functions


gongAll:
On the file input-types.ts ([link](https://github.com/fullcalendar/fullcalendar/blob/f5470c2cb63579b737172eca046196e232474f2e/packages/core/src/types/input-types.ts#L135)) seems to me that the type of validRange is not correct, it should be as visibleRange, per documentation ([link](https://fullcalendar.io/docs/validRange)).

acerix:
Would you be able to post a [runnable, stripped-down demonstration of the bug](http://fullcalendar.io/wiki/Reduced-Test-Cases/) or explain what is not working? Would really appreciate it because the time saved reproducing will be time spent fixing.

gongAll:
Why was it closed? Don't understand what demonstration is needed...

acerix:
I'm not sure why the definition is not correct, does it cause an error?

gongAll:
@acerix if you define a callback function as the argument, it will give you an IDE error (not that much of a big deal, I know, it was just a small change), but according to the documentation we are supposed to be able to use a callback function, and sure enough, if I do use, it will work, but all the method will be highlighted as red because it doesn't respect the definition of validRange. It's just a small bug, I believe.

I believe the change would be change this:
`
  validRange?: DateRangeInput
`

to this:

`
  validRange?: ((currentDate: Date) => DateRangeInput) | DateRangeInput
`
as per documentation.

acerix:
Thank you for explaining, that makes sense

arshaw:
this has been fixed in v5

PR #5372, opened 2020-04-24
Close #5182 and #5287

### LINKED ISSUE #5182: Event source function successCallback only accepts EventInput type, should allow anything because of eventDataTransform
I'm using event source function to fetch the events from API. Then to transform that events to match the fullcalendar model I'm using `eventDataTransform`. 

The problem is that I need to cast my model to `EventInput` before passing it to `successCallback` as it's not compatible. 

Wouldn't it be good if `successCallback` could receive `EventInput[] | any[]` (or something similar)?
I know I may do the model conversion inside custom event function, but why is the `eventDataTransform` called for that events?

gongAll:
Small corrections to types. 
#5182
#5287

### DIFF
--- a/packages/core/src/event-sources/func-event-source.ts
+++ b/packages/core/src/event-sources/func-event-source.ts
@@ -12,7 +12,7 @@ export type EventSourceFunc = (
     endStr: string
     timeZone: string
   },
-  successCallback: (events: EventInput[]) => void,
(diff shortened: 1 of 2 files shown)


arshaw:
this has been fixed in v5. thanks anyway for the PR @gongAll

PR #5391, opened 2020-05-03
Add support for jQuery 3.5

cedk:
In 3.5, jQuery changed the jQuery.htmlPrefilter to no more
XHTML-compliant. So tags must be closed for HTML.

### DIFF
--- a/src/Toolbar.ts
+++ b/src/Toolbar.ts
@@ -39,7 +39,7 @@ export default class Toolbar {
       el.append(this.renderSection('left'))
         .append(this.renderSection('right'))
         .append(this.renderSection('center'))
-        .append('<div class="fc-clear"/>')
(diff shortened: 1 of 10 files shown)


acerix:
Thanks, here's the issue for reference https://github.com/fullcalendar/fullcalendar/issues/5348

SudharsanSukumar:
Thanks for getting this fix out! Any ballpark estimate for when the fix will be available?

imouaddine:
@cedk thanks for the PR! @acerix when do you think approximately the fix would be available? we want to upgrade jQuery to 3.5.0 in our app, but fullCalendar is the only blocker. Thank you so for everything!

arshaw:
fixed. see my comment here: https://github.com/fullcalendar/fullcalendar/issues/5348#issuecomment-630563193

thank you for the help!

PR #5401, opened 2020-05-08
Add missing keys to EventSourceFunc fetchInfo definition

hmnd:
In addition to title, extracts fetchInfo into exported interface and renames generic "arg" to "fetchInfo" to match the docs.


acerix:

hmnd:
@acerix I'm not quite sure what I would demonstrate? All this does is add a few missing keys to the event function TS definition...

acerix:
Please open an issue describing the problem (eg. do you get an error when building?)  since we use issue for tracking and not PRs.

arshaw:
this is no longer an issue in v5
https://fullcalendar.io/blog/2020/06/v5-official

PR #5410, opened 2020-05-12
allow object in JSON feed

xi:
[Top level JSON arrays](https://stackoverflow.com/questions/3503102/) have often been considered insecure. I am not sure if this is still the case or if all browsers have been patched. Still I think it is better to err on the side of caution.

So this adds an option to wrap JSON feeds in an object.

### DIFF
--- a/packages/__tests__/src/legacy/events-json-feed.js
+++ b/packages/__tests__/src/legacy/events-json-feed.js
@@ -200,4 +200,26 @@ describe('events as a json feed', function() {
     expect(currentCalendar.getEventSources()[0].url).toBe('my-feed.php')
   })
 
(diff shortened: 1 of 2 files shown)


acerix:
Please submit this ticket under Issues with a [runnable, stripped-down demonstration](http://fullcalendar.io/wiki/Reduced-Test-Cases/) or what is not working, then link this PR to it.  

Custom JSON formats are supported by `eventDataTransform`:

https://fullcalendar.io/docs/eventDataTransform

xi:
Thanks for the feedback. https://fullcalendar.io/docs/eventSourceSuccess did the trick.

PR #5729, opened 2020-08-17
Fix state parameter of loading events

px-stkn:
Use the loadingLevel of the final state, to trigger the loading
callback with the correct isLoading parameter value.

A calendar with resources would emit a sequence of

  loading(): isLoading: true
  loading(): isLoading: false
  loading(): isLoading: false
  loading(): isLoading: false

before applying this patch. With this change applied, the
sequence turns into the expected:

  loading(): isLoading: true
  loading(): isLoading: false
  loading(): isLoading: true
  loading(): isLoading: false

Codepen for FC 5.3 before applying this patch:

   https://codepen.io/px-stkn/pen/KKzMaMj

after applying this patch:

   https://codepen.io/px-stkn/pen/xxVOgpd

The reduced testcase behaves slightly different than our (more complex) scenario, but it still
emits a "true, false, false" sequence of events.

### DIFF
--- a/packages/common/src/reducers/CalendarDataManager.ts
+++ b/packages/common/src/reducers/CalendarDataManager.ts
@@ -288,9 +288,9 @@ export class CalendarDataManager {
     }
 
     // TODO: use propSetHandlers in plugin system
-    if (!prevLoadingLevel && loadingLevel) {
+    if (!prevLoadingLevel && newState.loadingLevel) {
       emitter.trigger('loading', true)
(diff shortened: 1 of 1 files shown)


acerix:

arshaw:
solved via https://github.com/fullcalendar/fullcalendar/commit/386b30301b44feac6e387fe0c0ae71a76693c30c

arshaw:
this is now fixed in 5.5.0

PR #4674, opened 2019-05-08
Dragged element in wrong position when parent have css transform property

Calvin-LL:
Fix for issue #4673

Issue #4673, opened 2019-05-08
Dragged element in wrong position when parent have css transform property


Calvin-LL:
### Bug Reports
When the parent element has css transform property, dragging events in the calendar shows incorrect position of the mirrored event.

Shown in: https://codepen.io/coolcyberbrain/pen/oRjaLv

CoolCyberBrain:
I think a good solution would be to add the mirrored event to document root instead of the root of the calendar

acerix:
Thanks, confirmed

numsu:
Is there a timeline for this? We're facing the same issue.

manukieli:
I've the same issue.
Is there any solution already?
Thank you.

arshaw:
The codepen link isn't working (points to github). Can you repost @CoolCyberBrain

CoolCyberBrain:
Oops didn't do the markdown correctly so when it's clicked it goes to github. The URL is correct, I just changed the markdown, should work now. If it doesn't then try copy pasting on to url bar. https://codepen.io/coolcyberbrain/pen/oRjaLv

arshaw:
In `master` I've created a new settings called `fixedMirrorParent` that will offer a workaround for this issue. I'll release very soon. I'll comment here when I do.

arshaw:
I've released the new setting:
https://fullcalendar.io/docs/fixedMirrorParent

Demo if it working in 5.4.0
https://codepen.io/arshaw/pen/LYZMJMp

pixel9:
Thank you! I think you meant to point to a different codepen (it looks like a reproduction of the bug instead) but I'll take a look at the update and the new API.

As always, I appreciate all the hard work!

arshaw:
Sorry about that. Here's the corrected pen: https://codepen.io/arshaw/pen/ZEOVMVw

FUGIFA:
@arshaw is there any solution for this issue in version 4?

PR #5807, opened 2020-09-09
Fix event position calculation when calendar is inside a translated/scrolling container

pixel9:
Fixes: https://github.com/fullcalendar/fullcalendar/issues/4673
Fixes: https://github.com/fullcalendar/fullcalendar/pull/4674

**Description**
Dragged calendar event can get severely misaligned when the calendar element is inside a transformed element. This makes moving events around very confusing especially when the calendar is placed within complex layouts.

**Reduced Test Case**
Reposting test case from issue 4673: https://codepen.io/coolcyberbrain/pen/oRjaLv

**Steps to Reproduce**
Render calendar inside a CSS translated container. The situation becomes worse if one of the offset parents is scrolled as well.

**Analysis**
The current implementation calculates position based on only the current offset parent of the event being dragged. This means any CSS transforms/scrolling of parent elements are unaccounted for.

A new utility function `getClientPosition` has been created that will recursively collect all the offsets/scroll position to calculate a more accurate position for cursor tracking. This utility function will handle any number of parent elements with CSS positioning/scrolling so should be much more robust in more complex layout scenarios. I wasn't sure if this function would be useful anywhere else so I opted to keep it in the same module rather than extract it into some place like `/common/util/`

### DIFF
--- a/packages/interaction/src/dnd/ElementMirror.ts
+++ b/packages/interaction/src/dnd/ElementMirror.ts
@@ -1,4 +1,4 @@
-import { removeElement, applyStyle, whenTransitionDone, Rect } from '@fullcalendar/common'
+import { removeElement, applyStyle, whenTransitionDone, Point } from '@fullcalendar/common'
 
 // Returns the left/top offset of an element relative to the document body
 function getClientPosition(el) {
@@ -31,7 +31,7 @@ export class ElementMirror {
   deltaY?: number
   sourceEl: HTMLElement | null = null
   mirrorEl: HTMLElement | null = null
-  sourceElRect: Rect | null = null // screen coords relative to viewport
+  sourceElPosition: Point | null = null // screen coords relative to viewport
 
   // options that can be set directly by caller
   parentNode: HTMLElement = document.body
@@ -40,7 +40,7 @@ export class ElementMirror {
 
   start(sourceEl: HTMLElement, pageX: number, pageY: number) {
     this.sourceEl = sourceEl
-    this.sourceElRect = getClientPosition(sourceEl)
+    this.sourceElPosition = getClientPosition(sourceEl)
     this.origScreenX = pageX - window.pageXOffset
     this.origScreenY = pageY - window.pageYOffset
     this.deltaX = 0
@@ -98,15 +98,15 @@ export class ElementMirror {
 
   doRevertAnimation(callback: () => void, revertDuration: number) {
     let mirrorEl = this.mirrorEl!
-    let finalSourceElRect = getClientPosition(this.sourceEl) // because autoscrolling might have happened
+    let finalSourceElPosition = getClientPosition(this.sourceEl) // because autoscrolling might have happened
 
     mirrorEl.style.transition =
       'top ' + revertDuration + 'ms,' +
       'left ' + revertDuration + 'ms'
 
     applyStyle(mirrorEl, {
-      left: finalSourceElRect.left,
-      top: finalSourceElRect.top
+      left: finalSourceElPosition.left,
+      top: finalSourceElPosition.top
     })
 
     whenTransitionDone(mirrorEl, () => {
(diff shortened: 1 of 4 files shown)


arshaw:
Hi @pixel9, thanks so much for taking the time to write this PR. Before critiquing the code itself, I want to challenge some of the assumptions you're making:

> The current implementation calculates position based on only the current offset parent of the event being dragged

As it stands, the offset is being queries with [getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect), which returns results relative the the viewport's top/left. It does not consider any other "offset parents" when doing this.

> This means any CSS transforms/scrolling of parent elements are unaccounted for.

That's not true, `getBoundingClientRect` DOES account for transformations on parent elements. Please see this demo: https://codepen.io/arshaw/pen/Vwaxgjd ... the "inner" element is analogous to an event element.

I'm not saying your PR doesn't work, I'm just saying it might accidentally work for reasons you don't realize.

arshaw:
The bug is happening because of some counterintuitive behavior with fixed-positioned elements (example: events while being dragged) living WITHIN transformed elements. 

https://developer.mozilla.org/en-US/docs/Web/CSS/position:

> fixed - The element is removed from the normal document flow, and no space is created for the element in the page layout. It is positioned relative to the initial containing block established by the viewport, **except when one of its ancestors has a transform, perspective, or filter property set to something other than none** (see the CSS Transforms Spec), in which case that ancestor behaves as the containing block. (Note that there are browser inconsistencies with perspective and filter contributing to containing block formation.) Its final position is determined by the values of top, right, bottom, and left.

The bug is occurring because FullCalendar has not accounted for this.

Please see this illustration:

https://codepen.io/arshaw/pen/qBZYgap

arshaw:
In my opinion, the solution is not to query for the `sourceEl`'s position differently, it's to assume a different origin when SETTING the mirror's top/left if it turns out that it lives within a transformed ancestor.

packages/interaction/src/dnd/ElementMirror.ts
```diff
@@ -1,4 +1,21 @@
-import { removeElement, applyStyle, whenTransitionDone, Rect } from '@fullcalendar/common'
+import { removeElement, applyStyle, whenTransitionDone, Point } from '@fullcalendar/common'
+
+// Returns the left/top offset of an element relative to the document body
```
arshaw, line 3: Considering that the previous way of computing the position, `getBoundClientRect`, returns results relative to the viewport, and now this function returns them relative to the `<body>`, I would expect your new implementation to cause bugs for users who are NOT using transforms at all but who ARE scrolled down on their page.

pixel9: Thank you for taking a closer look. I hadn't considered the inverse in regards to scrolling without transforms present, but I can confirm that this works when both are present. I'll dig a bit deeper to make sure there aren't any issues.

pixel9:
> In my opinion, the solution is not to query for the `sourceEl`'s position differently, it's to assume a different origin when SETTING the mirror's top/left if it turns out that it lives within a transformed ancestor.

Wouldn't this require traversing ancestors anyway to determine this? I feel it's simpler to calculate a position using a consistent algorithm instead of adding branching logic. I'm not presuming my algorithm is bulletproof though...

pixel9:
@arshaw I've updated with an alternative solution. Please let me know if you feel this is a more viable approach and if there's anything else I can do to help resolve this issue.

In a nutshell, I capture enough information to pass along the relative position of the target element instead of trying to work backwards inside of the mirror element. It may turn out that additional properties are not needed and the existing `deltaX/Y` properties can be updated to calculate differently. I wasn't familiar enough with the codebase to measure the potential impact of such a change so I added new properties for now to prove out the idea.

arshaw:
hey @pixel9 , thanks, i'll take a closer look early this week

arshaw:
i made my own solution to this which i will release soon. more info: https://github.com/fullcalendar/fullcalendar/issues/4673#issuecomment-722689544

PR #5962, opened 2020-11-10
Turning private extendedProperties visible

fariaseduv:
To solve problems like this https://stackoverflow.com/questions/63145351/how-to-get-all-of-the-google-calendar-extended-properties-for-an-event-with-jque


acerix:

arshaw:
I realized this bug existed while working on the next release and already fixed it (with my own approach). it is queued for release. thank for the PR

PR #5969, opened 2020-11-13
iCal review

njpearman:
@arshaw this should be everything for the iCal feature. Note that I had to recreate the `yarn.lock` file after rebasing from `master`. I've run a build and the tests, and everything is green. I've also rerun `yarn lint` prior to opening this PR. Let me know whether you want any further tweaks.


examples/icalendar.html
```diff
@@ -0,0 +1,69 @@
...
+<script src='../packages/bundle/main.js'></script>
+<script src='../packages/icalendar/main.global.js'></script>
+<script>
+
+  document.addEventListener('DOMContentLoaded', function() {
+    var calendarEl = document.getElementById('calendar');
+
+    var calendar = new FullCalendar.Calendar(calendarEl, {
+      events: {
+        feedUrl: '/examples/ical/feed.ics',
```
arshaw, line 16: Could you please change this to a relative URL... `'ical/feed.ics'` ?

My server's document root is higher up and it didn't work. I think a relative URL is more flexible in this case.

packages/icalendar/src/main.ts
```diff
@@ -0,0 +1,105 @@
...
+    return null
+  }
+
+  const fcEvent = {
+    title: event.summary,
+    start: event.startDate.toString(),
+    end: (event.endDate ? event.endDate.toString() : null),
+  }
+
+  try {
```
arshaw, line 55: There isn't anywhere within the body of this `try` statement where an ICAL-related error would be thrown. Did you mean to wrap a try/catch around the `new ICAL.Event(vevent)` statement higher up in the function?

packages/icalendar/src/main.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return fcEvent
+  } catch (error) {
+    console.warn(`Unable to process item in calendar: ${error}.`)
+    return null
+  }
+}).filter((item: EventInput | null) => item !== null)
+
+let eventSourceDef: EventSourceDef<ICalFeedMeta> = {
```
arshaw, line 70: When running your icalendar.html demo, I noticed that whenever the user navigates prev/next, a request is made to the ical feed. This is pointless because we've already concluded that ical feeds are static and don't change based on the current date range.

The fix for this should be simple. Could you add a `ignoreRange: true` statement, like what exists in the array data source:
https://github.com/fullcalendar/fullcalendar/blob/v5.4.0/packages/common/src/event-sources/array-event-source.ts#L6

Then, can you add a test to make sure the http-mocked feed doesn't get called when a `calendar.next()` call is made? https://fullcalendar.io/docs/Calendar-next

packages/__tests__/src/icalendar/day-view.ts
```diff
@@ -0,0 +1,169 @@
...
+  it('adds a one-hour long meeting', (done) => {
+    loadICalendarWith(oneHourMeeting, () => {
+      setTimeout(() => {
+        assertEventCount(1)
+        done()
+      }, 100)
+    })
+  })
+
+  xit('adds a repeating weekly meeting', (done) => {
```
arshaw, line 36: Could you try to get this test working? Not sure what the hurdle is... you know the `initialDate`, and you can get the underlying event data from `Calendar::getEvents`, right?

arshaw: I understand a little more now. Please don't care about what is rendered. Only care about what is returned from `Calendar::getEvents` in this case.

tsconfig.json
```diff
@@ -9,10 +9,12 @@
     { "path": "packages/common" },
     { "path": "packages/daygrid" },
     { "path": "packages/timegrid" },
+    { "path": "packages/icalendar" },
     { "path": "packages/interaction" },
     { "path": "packages/list" },
     { "path": "packages/bootstrap" },
     { "path": "packages/google-calendar" },
+    { "path": "packages/icalendar" },
```
HoldYourWaffle, line 17: Is it just me or was this line added twice?

packages/icalendar/src/main.ts
njpearman: 👍 I think this was left hanging after I moved some code around. I've moved it up to the creation of the `ICAL.Event`, as you indicated.

packages/__tests__/src/icalendar/day-view.ts
njpearman: 👍 I'll remove pending from the test and check that the number of events matches all that are included in the feed.

njpearman: My `assertCount` method actually already handles both cases, rendered and data, but I hadn't made the method signature flexible enough to handle a difference. I've improved the method so now I can test what I had originally intended.

arshaw:
this has been released in v5.5.0. See comment here: https://github.com/fullcalendar/fullcalendar/issues/1580#issuecomment-748535186

PR #6080, opened 2021-01-13
fix UNTIL at the end of rrule

MicleBrick:
Include \n in UNTIL regex exclusion. This fixes input strings such as:
```
DTSTART;TZID=America/New_York:20210112T080000
RRULE:FREQ=MONTHLY;UNTIL=20210110T070000
RDATE;TZID=America/New_York:20210112T080000
```
Without this fix, parsing this input string results in a null exception.


acerix:

Issue #6097, opened 2021-01-18
Return the event's description in iCalendar package


diep-vu:
### Checklist

  - [x] I've already searched through [existing tickets](https://fullcalendar.io/issues)
  - [x] Other people will find this feature useful

### Feature Description

I think I might be useful that iCalendar returns in the extendedProps field the event's description. I have tried to fork your project and add this extra field on event building but I am not able to generate the javascript file (main.global.js) with the global variable FullCalendarICalendar.

I am looking forward to having your answer.

Thank you

arshaw:
queued for release

arshaw:
this has been released!
https://github.com/fullcalendar/fullcalendar/releases/tag/v5.6.0

PR #6136, opened 2021-02-03
Feature: Add location, organizer and description from ICAL events

k-cybulski:
This pull request will load the event organizer, location and description from ICAL calendars when using them as an event source. They are inserted as attributes of event.extendedProps.

This should make some common ICAL usecases possible, and solves #6097


arshaw:
merged and queued for release

PR #6139, opened 2021-02-03
Fix ICAL recurring event start and end times

k-cybulski:
The starting time was defined in the iCalEvent.iterator call wrongly as UTC midnight, whereas it should be the iCalEvent.startDate (the default when no arguments are given).

For an example of this going wrong, see this calendar with some test recurring events: https://ical-cors-proxy.herokuapp.com/?https://use01.thegood.cloud/remote.php/dav/public-calendars/BHNK5Arjn4S7GMGX/?export

On the current master branch, the recurring event dates are all at UTC midnight. With this branch, they will correctly be the time as set in the ICAL.


arshaw:
queued for release

arshaw:
this is released!
https://github.com/fullcalendar/fullcalendar/releases/tag/v5.6.0

Issue #5024, opened 2019-09-06
Google Calendar Attachments


pearcemichal:
### Feature Requests
I've been trying to find an answer for this, and so far everything I've found indicates that it is currently impossible for fullcalendar to get event attachments from google calendar. However, all of the answers are from several years ago, and their reasoning is that the attachments feature is still in beta. 

Has there been any progress on this since than? according to the google [calendar APi documentation](https://developers.google.com/calendar/v3/reference/events), information about attachments can be sent with requests, so it should be possible to get that info using fullcalendar, right?

In short: is it possible to add support for google calendar attachments? 

Regardless, thanks for all the work thats been put into this, it's a great calendar framework!

acerix:
I can't easily test this, but you might find the attachments in `extendedProperties`.

https://fullcalendar.io/docs/google-calendar#extended-properties

Otherwise that would be a new feature, so for now you would need to query the API manually to get them.

pearcemichal:
The attachment does not appear to be in the extended properties.

shudas:
Hi not sure if this is still actively being looked at but I needed to get this info for a project I was working on as well. I found a simple solution that would put it on the extendedProperties on the Events https://github.com/fullcalendar/fullcalendar/pull/6158

arshaw:
I've merged https://github.com/fullcalendar/fullcalendar/pull/6158 and will release it in the next release. thanks!

mtelloivan:
Hola necesito darme de baja de la lista de github y no me deja loguearme en
la web me dice que mi correo ***@***.*** es inexistente,

[image: image.png]

Gracias

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Libre
de virus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

arshaw:
this is released!
https://github.com/fullcalendar/fullcalendar/releases/tag/v5.6.0

PR #6158, opened 2021-02-10
add google attachments to event extendedProps

shudas:
Addresses https://github.com/fullcalendar/fullcalendar/issues/5024

Turns out attachments are [returned on the event objects](https://developers.google.com/calendar/v3/reference/events) themselves by Google.

Adding them as properties to the rawEventDef makes them propagate to extendedProperties in the FullCalendar Event so this should be the only change necessary to allow us to have access to the same attachment info


shudas:
Hi any updates on getting this into the master branch?

PR #6207, opened 2021-03-22
fix Popover closing inside Shadow DOM

### LINKED ISSUE #6205: Clicking inside the popover closes it with shadow DOM
### Reduced Test Case
Popover will be closed after click inside popover of the more content.

https://cg0md.csb.app/
https://codesandbox.io/s/recursing-pascal-cg0md

### Bug Description
Popover will be closed after click inside (on event).  Problem will be fixed after change 
`ev.target` to `ev.path[0]` in https://github.com/fullcalendar/fullcalendar/blob/master/packages/daygrid/src/Popover.tsx#L70

belozer-ti:
fixes #6205

**Description**
Popover closing each time by click inside Popover. Problem actual when we using Fullcalendar inside the Shadow DOM. Click listener has an incorrect target because target equals shadow dom root element. Example: https://codesandbox.io/s/recursing-pascal-cg0md?file=/src/full-calendar-component.js.
 
Current PR fixes the problem via `composedPath()` https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath

### DIFF
--- a/packages/daygrid/src/Popover.tsx
+++ b/packages/daygrid/src/Popover.tsx
@@ -67,7 +67,8 @@ export class Popover extends BaseComponent<PopoverProps> {
     let { onClose } = this.props
 
     // only hide the popover if the click happened outside the popover
-    if (onClose && !this.rootEl.contains(ev.target)) {
+    const target = ev.composedPath?.()[0] ?? ev.target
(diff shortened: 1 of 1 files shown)


belozer-ti:
@arshaw ping

PR #6237, opened 2021-04-03
Fix RRule not processing RDATE param

minemalox:
Hey,

while working with the fullcalendar rrule implementation I found out that the "RDATE" param, while using an rrule string, is not being parsed, which makes fullcalendar interpret the date as the current Time Zone and not the provided one.


acerix:

arshaw:
yes, we'll need a ticket for this.

i looked into RDATE. apparently its a way for the ICS to hardcode occurrences dates.

it seems to accept *lists* of dates, so parsing it will not be as straightforward as using `processMatch`
https://www.kanzaki.com/docs/ical/rdate.html

PR #6252, opened 2021-04-09
Continuous Integration using Github Actions

Naguumo:
### Features

Move CI pipeline to Github Actions
Provides one workflow with 4 parallel jobs to lint and test every PUSH

### Changes

- [x] Removes .travis.yml (configuration for Travis CI)
- [x] Replaces badge on readme to point to Github Actions
- [x] Removes CI scripts from package.json since they are called directly
- [x] Adds .github/workflows/continuous-integration.yml (the Github Actions config)
- [x] Runs parallel CI jobs on PUSH
- [x] Caches yarn installs to speed up run time
- [x] Caches example-projects when necessary for CI

### Examples

Test runs of these changes can be found at my fork [Here](https://github.com/Naguumo/fullcalendar/actions/workflows/continuous-integration.yml)

image

image


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Check Dependencies
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: .yarn/cache
+          key: yarn-cache-${{ github.job }}
+          restore-keys: yarn-cache-
+      - name: Install Dependencies
+        run: yarn install
+      - run: yarn run gulp lint
```
arshaw, line 27: could we give this step a name? "Lint Source Files"

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Build Project
+        run: yarn run gulp build
+      - run: yarn run gulp lintBuilt
```
arshaw, line 30: can we give this a name too: "Lint Built Files"

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,137 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: "recursive"
+      - name: Setup Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: "13"
```
arshaw, line 17: I was wondering, could we change this to node 14? I'd like that to be the node version we settle on for all our environments, since it's the newest LTS release.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,137 @@
+name: Continuous Integration
+on: [push]
+env:
+  TZ: "America/Los_Angeles"
+jobs:
+  ci-lint:
```
arshaw, line 6: for this job and all jobs like it, can we remove the "ci" in the identifier and the name? so...
```
ci-lint  -->  lint
"CI Lint" --> "Lint"

arshaw (commented):
Looks great! There's a fair but of repeated config, but that seems inevitable. I bet it will become more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) when we figure out how to get branching/artifacts working in the future.

Naguumo:
I would also love to make it more DRY, moving the repetitive parts to a common job will be very high on the priority list when I revisit this. For now, I have made the changes requested and a run of the Github Actions for it can be found [here](https://github.com/Naguumo/fullcalendar/actions/runs/741159140).

arshaw:
merged via squash commit: https://github.com/fullcalendar/fullcalendar/commit/47d4de93d13a08a37c5a73f76765dd5b3b30f6ef

PR #6266, opened 2021-04-15
use <span> on Event when no  is provided

### LINKED ISSUE #5710: Interactive content inside eventContent hook produces invalid DOM
### Reduced Test Case

https://codepen.io/HarlesPilter/pen/jOqbwzB

### Bug Description

eventContent injects our custom content into the inner most wrapper of the event. But a few nodes up the tree we have a <a> tag (even if we don't use any dragging or click features). This way if we use any interactive elements in our eventContent, it creates invalid DOM. (https://html.spec.whatwg.org/#the-a-element Content model) and currently there is no way to get rid of the wrapping <a> tag.
This greatly limits the use for eventContent.

What would fix this is when we don't use any interactive features on the event(dragging etc), then instead of the <a> tag we render <div> or even better would be to allow to override the tag with an attribute "as"

06cf4ebb-94a2-4289-8903-5470edceaa55

brandonwestcott:
As described in #5710, current rendering produced invalid DOM when event render contains an `a` element. While `onClick` handlers inside the event allow a valid DOM, it limits the ability to have the normal benefits of an anhcor tag, including "Open Link in New Tab" and a11y compatibility.

This implements a conditional check to see if `url` is set on the `EventDef`, if present use `a`, else use `span`.

Now, one _could_ argue `a` => `span` | `a` could be considered a breaking change. If so, an alternative could be to add an additional attribute onto the `EventDef` to control the wrapping element type.

Resolves #5710


vrinceanuv:
Any updates on this?

eloyesp:
Just wanted to add that I think that it should be a `div` not a `span` by default, as the later will have have other nesting issues, as it will make it impossible to use block elements.

brandonwestcott:
Happy to rebase this if @arshaw gives a 👍 for the proposed change.

ben467834:
I think this is still an issue. I don't think we're expecting block elements here so span is still the best inline candidate to be an anchor alternative.

dc-devs:
Bumping as this issue looks like it hasn't been commented on in a while. I would also love to see this update. 

Couple ideas for other possible solutions:

1. Would the solution of adding a custom `eventTagType` for an event be possible? This would introduce a non breaking change, and allow us to customize the element..
```
<FullCalendar
 ...
eventTagType={'span'}
/>
```

2. Would we be able to add our own `eventComponent`, and not just have access to updating the `eventContent`?
<FullCalendar
 ...
eventComponent={<CustomEventComponent />}
/>
```

PR #6282, opened 2021-04-26
Example Projects Improvement

Naguumo:
# What is the purpose of this change?

This change was put in place to update the example-projects repo to be more up to date and streamline the process by which yarn PnP is used. Furthermore, it attempts to provide a path for linting each of the example projects.

# What does this change do?

- [x] Update the example-run script
- [x] Creates an example-lint script
- [x] Adds a .prettierrc (currently unused) and a .eslintrc for linting configuration
- [x] Creates (and disables) a parcel-2 example project
- [x] Migrates the vue-typescript and vue-vuex examples to vite which works with yarn PnP
- [x] Updates nuxt to use PnP webpack plugin
- [x] Reorganizes the workspaces within the project for a more streamlined experience

# What does this change not do?

- [ ] Create a proper eslint configuration compatible with each project as this is to be discussed
- [ ] Allow the parcel-2 project to run
- [ ] Allow the NextJS projects to run

### Please note this PR should be done with consideration of [this PR](https://github.com/fullcalendar/fullcalendar-example-projects/pull/21) in the example-projects repo

The CI Run for this PR can be seen [here](https://github.com/Naguumo/fullcalendar/actions/runs/786346647)


scripts/example-run.js
```diff
@@ -1,73 +1,77 @@
-const path = require('path')
-const exec = require('./lib/shell')
-const globby = require('globby')
+const path = require("path");
```
arshaw, line 1: it looks like your editor may have auto-modified semicolons and quotes. it's not really a big deal, but in the future, can you make sure your editor doesn't do this? it will be a different story when we have eslint/prettier checking these `scripts` files.

Can you change this script back to using *single quotes* and *no semicolons*?

scripts/example-run.js
```diff
@@ -1,73 +1,77 @@
...
-let givenProjName = process.argv[2]
-let runCmd = process.argv[3]
+const rootDir = path.resolve(__dirname, "..");
+const examplesDir = path.join(rootDir, "example-projects");
+const givenProjName = process.argv[2];
+const runCmd = process.argv[3];
+
+///////////////////////////////////////////////////////
+// Project Settings
+const redirectProjects = [["parcel", "parcel-2"]];
```
arshaw, line 12: Instead of having this `redirectProjects` config, can we simply consider `parcel` (parcel 1) and `parcel-2` as being disabled projects?

scripts/example-run.js
```diff
@@ -1,73 +1,77 @@
...
+const disabledProjects = [
```
arshaw, line 13: A more readable and more efficient way of defining `disabledProjects` would be to make a map of `{ '<project-name>': '<disabled-reason>' }`.

scripts/example-run.js
```diff
@@ -1,73 +1,77 @@
...
+  ["next", "This example is disabled till the next major release"],
+  ["next-scheduler", "This example is disabled till the next major release"],
+  [
+    "parcel-2",
+    "There is currently a bug in parcel bundler which prevents this from working",
+  ],
+];
+const pnpSimulatedProjects = ["angular"];
```
arshaw, line 21: Let's make this follow the new pattern for `disabledProjects` (a map of project-name:reason). The reason for angular using pnp should be "Angular CLI doesn't support Yarn PnP".

scripts/example-run.js
```diff
@@ -1,73 +1,77 @@
...
+  } else {
+    console.log("Normal Yarn execution");
   }
 
-  console.log('')
-}
+  // Execute
+  console.log();
+  exec.sync(["yarn", "run", ...execCmd], {
+    cwd: projDir,
```
arshaw, line 72: For yarn PnP simulation, the `cwd` (current working directory) is supposed to be the `rootDir` not the `projDir`. It seems like it's still working however, likely due to some yarn-run fallback behavior where if a run-command is not found, it searches parent directories. I'd rather not rely on this behavior however. It'd be more understandable to explicitly reference the root's run script.

So, it'd probably be best to have individual `exec.sync` calls for both the `if` and `else`. It would also eliminate the need for conditional `execCmd` assignment, which is not very readable IMO.

arshaw (changes requested):
Everything looks and works great! I just had some minor comments.

Naguumo:
@arshaw the changes you requested have been made. Please take note however that I have added a .prettierrc to the root of the project. While prettier is not being used as a dependency of the project this does allow IDEs to better understand the code style we are using and autocorrect on save for it. The way that we were attempting to enforce such behavior was through the .editorconfig but unfortunately it does not provide any support for semicolons. This addition is very much open for further discussion

scripts/example-run.js
```diff
@@ -17,57 +31,42 @@ if (!runCmd) {
...
-  console.log(projDir)
+const projNames =
+  givenProjName === 'all'
+    ? globby.sync('*', { cwd: examplesDir, onlyDirectories: true })
+    : [givenProjName]
 
-  switch(projName) {
+projNames.forEach((projName) => {
+  // Don't run disabled projects
+  if (Object.keys(disabledProjects).includes(projName)) {
```
arshaw, line 41: why not just `if (disabledProjects[projName])` ?

if you're concerned about making sure we correctly handle empty strings / undefined / nulls in the map's values, then `if (disabledProjects.hasOwnProperty(projName))` is good.

Converting an object's keys to an array is a bad habit to get into because it converts what could be *constant* time (the performance reason you use hashes) to *linear* time (searching through an array's items one-by-one)

scripts/example-run.js
```diff
@@ -17,57 +31,42 @@ if (!runCmd) {
...
+    const disabled = disabledProjects[projName]
```
arshaw, line 42: this line and the next are overly deconstructed. `console.info(disabledProjects[projName])` would be fine.  The only reason I could empathize with for breaking it into two lines is to explain the purpose of the hash values by using a meaningful variable name such as `disabledReason`. Your choice on approach.

arshaw (changes requested):

scripts/example-lint.js
```diff
@@ -0,0 +1,48 @@
...
+    : [givenProjName]
+
+projNames.forEach((projName) => {
+  const projDir = path.join(examplesDir, projName)
+
+  console.log()
+  console.info('PROJECT:', projName)
+  console.log(projDir)
+
+  const { success } = exec.sync(['yarn', 'run', 'lint'], {
```
arshaw, line 21: It's not the best idea to let a command run that you know might fail. It is slower and might mislead someone when they look at this script's output.

Can you check whether a given has a lint script by inspecting its package.json? You can simply `require` the json file and look at the `scripts` map. I'd put this into its own utility function (`doesPackageHaveLinting` ?) Also please consider that an example project might not have a `scripts` map.

scripts/example-lint.js
```diff
@@ -0,0 +1,48 @@
+const path = require('path')
+const exec = require('./lib/shell')
+const globby = require('globby')
+
+const rootDir = path.resolve(__dirname, '..')
+const examplesDir = path.join(rootDir, 'example-projects')
+const givenProjName = process.argv[2] || 'all'
```
arshaw, line 7: it's simpler to remove the inclusion of an `'all'` string and simply check for falsiness in the next expression.

scripts/example-lint.js
```diff
@@ -0,0 +1,48 @@
...
+
+    // Prettier is currently not part of any workspace, I will defer on adding it as a dependency
+    // exec.sync(["yarn", "run", "prettier", "--write", "./src"], {
+    //   cwd: projDir,
+    //   exitOnError: false,
+    //   live: true,
+    // });
+
+    exec.sync(
+      'yarn exec eslint src/ --fix-dry-run --config ../.eslintrc.json --ext .tsx,.ts,.jsx,.js',
```
arshaw, line 38: I'm seeing a bunch of errors that look like this:

```
10:3   error  'document' is not defined  no-undef
36:5   error  'setTimeout' is not defined  no-undef
47:3   error  'console' is not defined     no-undef
49:14  error  'Promise' is not defined     no-undef
```

Is there any way to tell eslint we are in a browser environment and to expect these globals?

Also, I'm seeing errors like this:

```
11:17  error  Parsing error: Unexpected token :
```

It seems that typescript parsing isn't set up.

Could you adjust your eslint config so these errors are no longer errors?

arshaw (commented):
I merged everything you had, but there are still a few improvements I'd like. Can you make a new PR addressing this feedback, based off the latest `master`?

arshaw:
One more thing. In the projects that now use Vite, when you run the `build` command for each, they produce files in the `dist` directory, but when you attempt to open up `dist/index.html` in a normal file browser, your web browser loads a page that is broken. That's because absolute URLs are being used to access to the assets like `/assets/index.7c10d6e4.js` which result in a broken link like `file:///assets/index.7c10d6e4.js`.

I know I didn't initially say this as a requirement, but can we allow the `index.html` files to be runnable as-is, without a web server? It's probably just a matter of adding an assets url base of `./`

Issue #3396, opened 2016-10-28
Event overlaped with slotEventOverlap set to false


oscamen:
Sometimes events are overlaped with `slotEventOverlap: false` option.

Demo: https://jsfiddle.net/b14hkwun/2/

![demo](https://cloud.githubusercontent.com/assets/11599942/19803173/8e91c0ce-9d07-11e6-9b93-c05e2623e137.png)

arshaw:
dup of:
https://github.com/fullcalendar/fullcalendar/issues/2758
&uarr;
@oscamen, could you please visit the link, **give a** 👍  **to vote** ([?](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments)), press the 🔈**Subscribe** button to receive updates?

Issue #2758, opened 2015-08-19
timeGrid events incorrectly overlapping


arshaw:
Originally reported on Google Code with ID 2493

```

I found an unexpected scenario on fullcalendar where events start hiding each other
when many are on the screen. The minimal set I found to reproduce this error is composed
by nothing less than 12 events, and if any of these 12 events is removed or changed,
the bug disappears.

Please take a look on the demo at http://jsbin.com/xizujo/1/edit?js,output (or http://plnkr.co/F2bzehY8Pc9Vo3D4H5Lc)
and search for the events L and B. You will actually not be able to find B because
L is hiding it, so move L far from its position and B is gonna appear. The screenshots
1 and 2 show this process.

Does anybody has any clue what is going on with event B?
```

Reported by `vd@churchdesk.com` on 2015-04-28 10:01:28

<hr>
- _Attachment: 1.png<br>![1.png](https://storage.googleapis.com/google-code-attachments/fullcalendar/issue-2493/comment-0/1.png)_
- _Attachment: 2.png<br>![2.png](https://storage.googleapis.com/google-code-attachments/fullcalendar/issue-2493/comment-0/2.png)_

Imported with **2** stars.

arshaw:
Depiction of `vd@churchdesk.com`'s JSBin from a while ago:
![screen shot 2017-04-27 at 6 20 50 pm](https://cloud.githubusercontent.com/assets/56520/25506776/4f400e78-2b76-11e7-8755-6475fc7de6db.png)
http://jsbin.com/xizujo/2/

arshaw:
I'm not sure when this was fixed, but it's fixed now (as of v5.5.1)

https://codepen.io/arshaw/pen/ExZKNze?editors=0010
(notice how event "B" is now visible)

PR #6287, opened 2021-04-29
Fix bug with hidden/overlapped events in V3

kylethielk:
Hi There,

Not sure if you are still accepting PR for v3. We are on V3 for the foreseeable future due to a very tough upgrade for our particular use case.

We encountered the same bug as found in #2758 and #3396 and it became painful enough we took it upon ourself to find a solution. Turns out the best solution is the easiest one. This bug is not present in latest version of 5.X and thankfully this portion of the code has not changed much so it made comparing v3 to v5 incredibly easy and we were able to extract the following changes directly from 5.x to solve the issue.

The Bug as present in the latest v3.10.2: https://replit.com/@kylethielk/FullCalendarOverlapBug#index.html
The same code with the proposed fixes: https://replit.com/@kylethielk/FullCalendarOverlapBug-Fixed#index.html

Notice that in the first example the event "B" is not visible as it is hidden by "L".

### DIFF
--- a/src/agenda/TimeGridEventRenderer.ts
+++ b/src/agenda/TimeGridEventRenderer.ts
@@ -154,6 +154,14 @@ export default class TimeGridEventRenderer extends EventRenderer {
     let level0
     let i
 
+    // IMPORTANT TO CLEAR OLD RESULTS :(
+    for (i = 0; i < segs.length; i++) {
+      let seg = segs[i]
+      seg.level = null
+      seg.forwardCoord = null
+      seg.backwardCoord = null
+      seg.forwardPressure = null
+    }
     this.sortEventSegs(segs) // order by certain criteria
     levels = buildSlotSegLevels(segs)
     computeForwardSlotSegs(levels)
@@ -183,7 +191,7 @@ export default class TimeGridEventRenderer extends EventRenderer {
     let forwardSegs = seg.forwardSegs
     let i
 
-    if (seg.forwardCoord === undefined) { // not already computed
+    if (seg.forwardCoord == null) { // not already computed
 
       if (!forwardSegs.length) {
 


kylethielk:
This has caused some unintended consequences with the layout engine/algorithm incorrectly arranging events, particularly on days that have a large number of events and a significant portion of those events are long in duration. I'm closing this until we have time to investigate further.

arshaw:
fwiw, i completely rewrote this logic, fixing a log of bugs in the process, and i'll release it in a few days, but only for v5 :(

kylethielk:
Thank you for the comment @arshaw, good to know and I'll keep an eye out. If it does cause too many issues for us moving forward, I might try and port over the rewritten logic myself to V3.

If anyone stumbles on this and is looking for a quick and dirty solution, we are currently working around this with a jerry-rigged solution:

```javascript

var fc_computeFgSegHorizontals = FullCalendar.TimeGridEventRenderer.prototype.computeFgSegHorizontals;
FullCalendar.TimeGridEventRenderer.prototype.computeFgSegHorizontals = function(segs)
{
    fc_computeFgSegHorizontals.call(this,segs);
    /**
    * Find any segments that have a <= 0 width, increase their level by 1 (for unknown reason the hidden 
    * element will also be 1 z-index level below the item occupying the same space and thus we have to 
    * bump up level to get it to display on top) and set a somewhat sane default length of 8.67% of 
    * day width.
    */
    for(i = 0; i < segs.length; i++)
    {
        var seg = segs[i];
        if(seg.forwardCoord <= seg.backwardCoord)
        {
            seg.forwardCoord = 1-seg.backwardCoord + 0.0867;
            seg.level++;
        }
    }
};
 
```

Issue #5572, opened 2020-07-01
Add Welsh locale (cy-GB)


LeaTark:
Welsh is a legal requirement for public sector sites in wales.

I thought I'd be able to add the locale myself.

Apologies if I can, I just haven't figured out how yet.

Thank you.

acerix:
To add the locale, you need to create a file in this directory:

 https://github.com/fullcalendar/fullcalendar/tree/master/packages/core/src/locales

After adding the file to the GitHub repo you can submit it as a PR.

https://help.github.com/articles/using-pull-requests/

LeaTark:
I can't see where to provide translations for month and day names in those files.

I had a separate issue open for month and day names #5565 but it was closed.

acerix:
The day/month names are translated by the browser so are no longer defined in the locale files.

pvanthony:
I added a ta.js file for Tamil language but not sure how to do the PR.

PR #6400, opened 2021-06-24
Add Kurdish (Central Kurdish)

hawkarOagaz:
This file contains a Kurdish (Central Kurdish) translation. However, monthNames and dayNames are not defined here. Therefore, they are not added. This problem was mentioned in issue "Add Welsh locale (cy-GB) #5572" as well but still don't know how to fix it.


arshaw:
released in [5.9.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v5.9.0)

Issue #6445, opened 2021-07-09
Typing of `event.setProp` does not allow booleans


turisap:
### Reduced Test Case

[REDUCED TEST CASE](https://codesandbox.io/s/setprop-typing-628ng?file=/src/calendar.tsx).

### Bug Description
As per docs, we can set any non-date related props via this function. However, its typing states that it only accepts strings as values. Consequently, I get a TS error when trying set `event.setProp('editable', true)`

```
declare class EventApi {
    // ....
    setProp(name: string, val: string): void;
```

arshaw:
released in [5.9.0](https://github.com/fullcalendar/fullcalendar/releases/tag/v5.9.0)

PR #6446, opened 2021-07-09
Add boolean type to setProp value param

turisap:
Adds possibility to pass `boolean` values to `event.setProp` without TS errors.

resolves [#6445](https://github.com/fullcalendar/fullcalendar/issues/6445)


arshaw:
I've actually already resolved this with https://github.com/fullcalendar/fullcalendar/commit/5f698f1f2517cbcb4269c1a984f8520da3e1c04f

Rather accept an `any` type, because options could be many things.

Will release soon.

Issue #6448, opened 2021-07-11
Touch device, select not working when inside webcomponent


Bakino:
### Reduced Test Case

https://codepen.io/olibakino/pen/eYWdbRV?editors=1010

### Bug Description

In a touch device, when fullcalendar is included in a WebComponent, select an event does not work. The event is selected but immedialty unselected

It should remain selected like when fullcalendar is not included in a WebComponent

acerix:
I believe this existing ticket captures this need:
https://github.com/fullcalendar/fullcalendar/issues/6429
↑
Please visit the above issue, **give a** 👍  **to vote** ([?](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments)), press the 🔈**Subscribe** button to receive updates.

PR #6449, opened 2021-07-11
Fix select bug on touch device when inside a WebComponent

Bakino:
This change fix the bug #6448

Because the unselect check the mousedown event is the event element here : 
```javascript
this.matchesEvent = !!elementClosest(downEl, EventDragging.SELECTOR) // interaction started on an event?
```
but in WebComponent context, if we use the main document, the target of the event is the WebComponent, not the event el which is inside.

Changing to getRootNode() on the calendar element fixes the bug because it return the "document inside WebComponent" when running inside a WebComponent or the "normal document" in classical case

### DIFF
--- a/packages/interaction/src/interactions/UnselectAuto.ts
+++ b/packages/interaction/src/interactions/UnselectAuto.ts
@@ -14,7 +14,7 @@ export class UnselectAuto {
   matchesEvent = false
 
   constructor(private context: CalendarContext) {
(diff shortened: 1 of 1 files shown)


arshaw:
I've fixed this bug via https://github.com/fullcalendar/fullcalendar/commit/30cbd436e76ba51765c2cdc85967e7ef81f7352b
queued for release

Issue #6460, opened 2021-07-15
npm Package dependency issue with "jquery-simulate": "github:jquery/jquery-simulate", v3.10.2


MasterTBUUK:
### Reduced Test Case

not reproducible with a link

### Bug Description
using:
NODE v4.2.6
npm v3.5.2
package.json

{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "dependencies": {
    "fullcalendar": "3.10.2",
    "jquery": "3.5.1"
  },
  "devDependencies": {},
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC"
}
npm install

When trying to install this package from NPM the following error gets returned

fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
npm ERR! git rev-list -n1 master: Use '--' to separate paths from revisions, like this:
npm ERR! git rev-list -n1 master: 'git <command> [<revision>...] -- [<file>...]'
npm ERR! git rev-list -n1 master:
npm ERR! git rev-list -n1 master: fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
npm ERR! git rev-list -n1 master: Use '--' to separate paths from revisions, like this:
npm ERR! git rev-list -n1 master: 'git <command> [<revision>...] -- [<file>...]'
Cloning into bare repository '/root/.npm/_git-remotes/git-github-com-jquery-jquery-simulate-git-02e2b007'...
npm ERR! git clone --template=/root/.npm/_git-remotes/_templates --mirror git@github.com:jquery/jquery-simulate.git /root/.npm/_git-remotes/git-github-com-jquery-jquery-simulate-git-02e2b007: Permission denied (publickey).
npm ERR! git clone --template=/root/.npm/_git-remotes/_templates --mirror git@github.com:jquery/jquery-simulate.git /root/.npm/_git-remotes/git-github-com-jquery-jquery-simulate-git-02e2b007: fatal: Could not read from remote repository.
npm ERR! git clone --template=/root/.npm/_git-remotes/_templates --mirror git@github.com:jquery/jquery-simulate.git /root/.npm/_git-remotes/git-github-com-jquery-jquery-simulate-git-02e2b007:
[...]

PR #6461, opened 2021-07-15
V3

MasterTBUUK:


MasterTBUUK:
this was raised for issue #6460, to allow any npm to use the package without failing,

PR #6489, opened 2021-07-29
fix issue with creatating date from object

freoff:
Fix issue described here https://github.com/moment/luxon/issues/996


acerix:

arshaw:
I believe this was fixed when the luxon2 connector was implemented

Issue #5503, opened 2020-06-16
processing css with postcss, calc/var problems


adenta:
I do not know where to best report this, but it needs to be documented somewhere.

I am using a pretty standard [webpacker](https://github.com/rails/webpacker) configuration to run my react app that uses Fullcalendar. After upgrading to the V5 release candidate, I encountered the following error: 
```
ERROR in ./node_modules/@fullcalendar/timegrid/main.css (./node_modules/css-loader/dist/cjs.js??ref--7-1!./node_modules/postcss-loader/src??ref--7-2!./node_modules/@fullcalendar/timegrid/main.css)
Module build failed (from ./node_modules/postcss-loader/src/index.js):
ParserError: Syntax Error at line: 1, column 45
    at /Users/andre/terusama/node_modules/@fullcalendar/timegrid/main.css:84:7
    at Parser.error (/Users/andre/terusama/node_modules/postcss-values-parser/lib/parser.js:127:11)
    at Parser.operator (/Users/andre/terusama/node_modules/postcss-values-parser/lib/parser.js:162:20)
    at Parser.parseTokens (/Users/andre/terusama/node_modules/postcss-values-parser/lib/parser.js:245:14)
    at Parser.loop (/Users/andre/terusama/node_modules/postcss-values-parser/lib/parser.js:132:12)
    at Parser.parse (/Users/andre/terusama/node_modules/postcss-values-parser/lib/parser.js:51:17)
    at parse (/Users/andre/terusama/node_modules/postcss-custom-properties/index.cjs.js:47:30)
    at /Users/andre/terusama/node_modules/postcss-custom-properties/index.cjs.js:333:24
    at /Users/andre/terusama/node_modules/postcss/lib/container.js:190:18
    at /Users/andre/terusama/node_modules/postcss/lib/container.js:135:18
    at Rule.each (/Users/andre/terusama/node_modules/postcss/lib/container.js:101:16)
```

I was incredibly confused as to what to do about this. I understand that the new version of fullcalendar required a css loader, but it was my understanding that _I was already doing this_.

I was able to get webpack(er) to compile successfully by removing all uses of `calc` in Fullcalendar css manually. This is not a real solution, however; so I had to keep digging.

I was able to resolve my issues, by changing my postcss configuration from :
```
module.exports = {
  plugins: [
    require('postcss-import'),
    require('postcss-flexbugs-fixes'),
    require('postcss-preset-env')({
      autoprefixer: {
[...]

PR #6563, opened 2021-09-27
Fixed css postprocess build issue

rganeyev:
Context:
https://github.com/fullcalendar/fullcalendar/issues/5503#issuecomment-645654732

See https://developer.mozilla.org/en-US/docs/Web/CSS/calc
"The + and - operators must be surrounded by whitespace."

It works fine on live, but on postprocessing spaces are erased, leading to the error.

I replaced `... / -2` by `-.5 * ....`. Mathematically they are the same, but calc is happier


acerix:
Please submit this ticket under Issues with a [runnable, stripped-down demonstration](http://fullcalendar.io/wiki/Reduced-Test-Cases/) of the error, then link this PR to it.

rganeyev:
Resolves #6564

Issue #6593, opened 2021-10-14
Ability to be called back when the more popover window is closed


yenoiwesa:
### Checklist

Please mark these items with an [x]

  - [x] I've already searched through [existing tickets](https://fullcalendar.io/issues)
  - [x] Other people will find this feature useful

### Feature Description

Adding an option to map a callback that gets called when the more popover window is closed. Following the naming convention used by similar event mapping in FullCalendar, the new option would be called `morePopoverClose`.

### Context

Our team needs this information because we alter some functionality around the FullCalendar instance when the popover is open, and we need to be able to tell when it is closed so the functionality can be reverted back. At the moment there is no such option.

I have created a pull request for this feature suggestion.

Thanks!

carlababa:
Our team needs this functionality too. 
We've been avoiding using the default FullCalendar popover, because we need to trigger an action when the popover closes, and right now is not possible. 
The callback onClose would solve this for us.

PR #6594, opened 2021-10-14
Offer new morePopoverClose callback option

yenoiwesa:
Adding an option to map a callback that gets called when the more popover window is closed.
The new option is named `morePopoverClose`.

Feature request #6593


JoabeFerri:
Would you have a function that makes the popover close when you click on the meeting? @acerix

Issue #6683, opened 2022-01-07
iCalendar format seems to be ignored for event sources as a function


stshort:
### Checklist

Please mark these items with an [x]

  - [x] I've already searched through [existing tickets](https://fullcalendar.io/issues)
  - [x] Other people will find this feature useful

### Feature Description

I'm trying out the new ics support, however it seems like the implementation does not allow for passing access tokens properly in GET requests (I tried passing `access_token` and `token_type` in the URL, but to no success). So to get around this, I'm using event sources as a function. By doing this:

```
{
  events: EventSourceFunc[],
  format: 'ics',
}
```

I am able to fetch all of my events fine, however when I pass the ics file contents to the `successCallback`, it doesn't seem like they're being parsed, and will rely on me manually parsing the ics file and translating them to raw events. I'd expect if I am passing `ics` as the format, that it should hook into the same workflow as if I were to hook it into an ics feed as documented.

I also noticed that it seems to be doing the same non-ics behavior of fetching new data every time the user changes the calendar view or changes the calendar page. I'd expect that if I am using ics as a format, it should hook into the same logic, regardless of if I am using event sources as a function or a direct ics feed.

If I'm doing anything wrong I'd appreciate feedback, but from everything I can tell, this seems like it is likely a gap in the iCalendar implementation.

stshort:
Alternatively, if you can add support to customize the headers for ics/json feeds that would work too. As far as I can tell, that's not supported still

https://github.com/fullcalendar/fullcalendar/issues/4627

acerix:
Would you be able to post a [runnable, stripped-down demonstration of the bug](http://fullcalendar.io/wiki/Reduced-Test-Cases/)? Would really appreciate it because the time saved reproducing will be time spent fixing.

stshort:
@acerix I'm trying to gauge the difficulty of fixing it myself on a fork, and was looking through the code, I think the problem is:

https://github.com/fullcalendar/fullcalendar/blob/master/packages/common/src/structs/event-source-parse.ts#L93

In this event source parser, later added plugins are taking precedence. Which makes sense I suppose, however:

https://github.com/fullcalendar/fullcalendar/blob/master/packages/common/src/plugin-system.ts#L95

[...]

Issue #4627, opened 2019-04-10
How to set Authorization header in eventSources HTTP request


azuken:
Hello,

I'm trying today to migrate my company's app from v3 to v4, and I'm facing a problem with eventSources component. In v3, we used `beforeSend` function to get http request, and set Authorization header (for bearer token).

Now in v4, it does not call `beforeSend` function, and does not use `headers` params. I didn't find any doc or code from the internet with a solution. 

Is it just not implemented ? 

We've search in source and found nothing about it too.

What I tried :

(v3 base code)
```
beforeSend: function (xhr) {
    xhr.setRequestHeader('Authorization', 'Bearer ' + token);
},
```

(v4)
```
headers: { 'Authorization': 'Bearer ' + token }
```

```
headers: { Authorization: 'Bearer ' + token }
```

Thanks in advance !

acerix:
You could supply the events as a function which gives you more control: https://fullcalendar.io/docs/events-function

I believe it's not possible with "events as JSON" currently but we could look at adding that support if you are requesting that feature.

azuken:
Ok thanks for the link, I will look at it later. For the feature request, is there any procedure to do it ?

acerix:
This issue can be the "feature request" but we don't have an ETA for adding new features.  This page has the details:

https://fullcalendar.io/requesting-features

arshaw:
in v4, we should offer a hook to manipulate the XHR before it requests. keeping this ticket open for this feature request

MiguelFaria:
hi @arshaw, 

thare is any forecast about when this feature is gona be released ??

markoangelovski:
Dear @arshaw and team,

In addition to OP's note about setting the Authorization header, may I kindly add that it would be great if the provision was available for setting the xhr.withCredentials = true.

My project is using it for authorization purposes, and currently there is no way for setting this option without modifying the source.

Thank you.

m8iog:
I don't know if you've solved your problem @azuken, but I found a solution in the event source as a function.

                    events: function (info, successCallback, failureCallback) {
                        let start = moment(info.start.valueOf()).format('YYYY-MM-DD');
                        let end = moment(info.end.valueOf()).format('YYYY-MM-DD');
                        $.ajax({
                            url: "api/assignments/?event_id=" + vm.eventId +  '&start='+ start + "&end=" + end,
                            type: 'GET',
                            headers: {
                                'X-CSRF-TOKEN': window.csrf_token
[...]

PR #6690, opened 2022-01-09
Custom headers - Possible Implementation

stshort:
I gave a stab at implementing #4627 / updates for #6683 and this seems to be working for my use case now, and lets users pass additional headers to event source configurations. This lets you set Authorization headers for JSON / iCalendar feeds.

I didn't run the tests / update tests, because I'm simply trying this out on my project as this functionality being missing is a direct blocker to work that I am currently trying to implement. Pushing this PR so the maintainers can review and take if you would like, it really seems like an easy addition with a lot of added benefits.

Should work for JSON, Google Calendar, and iCalendar feeds, but I'm only using iCalendar on my project. I'm not familiar enough with the fullcalendar source to guarantee this change works across all event sources as intended, but it is now definitely letting me pass an auth header for an ics feed.

@acerix FYI

### DIFF
--- a/packages/common/src/event-sources/json-feed-event-source.ts
+++ b/packages/common/src/event-sources/json-feed-event-source.ts
@@ -34,7 +34,7 @@ let eventSourceDef: EventSourceDef<JsonFeedMeta> = {
   },
 
   fetch(arg, success, failure) {
-    let { meta } = arg.eventSource
+    let { meta, additionalHeaders } = arg.eventSource
     let requestParams = buildRequestParams(meta, arg.range, arg.context)
 
     requestJson(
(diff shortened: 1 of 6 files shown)


stshort:
@acerix @arshaw any updates / feedback on this implementation? Not being able to support authenticated endpoints for event source fetching seems like a big oversight.

I'll be using this implementation for my project but it would definitely be ideal for some implementation of this to be included in the native package.

If you are interested, I also needed to modify this library to support `DATE` type for `EXDATE` entries, and also include the iCalendar event UID to the extendedProps to easily identify which event the user is interacting with. I can raise a PR for this as well if you are interested.
https://github.com/stshort/fullcalendar/commit/4762611cbb106873f320e100c1bc84b570ad943e

It's a hack specifically for my project since I'm hardcoding the `onlyDate` param to always be true since I'm only using `DATE` types for exception dates, but the idea should be generalizeable and probably could even be configured as an event source option, or just simply support either/or.

Issue #6681, opened 2022-01-06
Support event color property for ICS calendars


materpillar:
### Checklist

Please mark these items with an [x]

  - [x] I've already searched through [existing tickets](https://fullcalendar.io/issues)
  - [x] Other people will find this feature useful

### Feature Description

With the just released [ical.js version 1.5.0](https://github.com/mozilla-comm/ical.js/releases/tag/v1.5.0), it is possible to read (and set) the color property for an event.
It would be very useful if fullcalendar was able to display the event using this color property. In fact, adding this feature
seems quite easy: All that is need to be done is to add: `color: iCalEvent.color,`
https://github.com/fullcalendar/fullcalendar/blob/ede23c4b2bf0ee0bb2cbe4694b3e899a09d14da6/packages/icalendar/src/main.ts#L147

Yet, I believe that there is an additional toggle option needed to chose if the color shall be used from the ics source, or if it shall be overwritten by the fullcalendar settings?

PR #6697, opened 2022-01-20
adding backgroundColor from iCalEvent

david-loe:
As Color is a valid iCalendar attribute ([rfc7986](https://datatracker.ietf.org/doc/html/rfc7986)), adding the event backgroundColor from iCalEvent


david-loe:
#6681

david-loe:
Is there a reason that this change was not merged jet?

captn3m0:
I'm generating calendars with multiple colors. Would be great if fullcalendar could support this so my calendar looks more colorful.

PR #6714, opened 2022-02-08
Use HTTPS url for vue 3 submodule

wartmanm:
Hi

I got stopped by a "permission denied (publickey)" error when following the setup instructions in CONTRIBUTING.md, and found that the cause was an SSH url used by the Vue 3 submodule.  I'm guessing this was an oversight since the other submodules all use HTTPS.  With this change, it's possible to clone the repository anonymously.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,4 +15,4 @@
 	url = https://github.com/fullcalendar/fullcalendar-example-projects.git
 [submodule "packages-contrib/vue3"]
 	path = packages-contrib/vue3
-	url = git@github.com:fullcalendar/fullcalendar-vue.git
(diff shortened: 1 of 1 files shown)


arshaw:
Queued for release
https://github.com/fullcalendar/fullcalendar-workspace/commit/44e836224c31079c7270347bc2288873d5d1c91c

The v6 version of the codebase has the git submodule root at this repo now:
https://github.com/fullcalendar/fullcalendar-workspace

I'm actually using relative URLs now, so if you clone the fullcalendar-workspace repo with https, all submodules will be cloned with https as well

arshaw:
Released in [v6.1.1](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.1)

PR #7018, opened 2022-11-02
proposal: add formatting feature to weekLabel

hrkt:
This PR is a proposal for weekLabel text formatting option for localization.

Motivation:

Current implementation of weekLabel can handle localization with a limiation of its label-and-number order.

i.e.
Week 1 (LTR)

Whereas in Japanese, week label is more suitable when it is rendered
number-and-label order, or label-number-label order.

i.e.
1週 (LTR)
第1週(LTR)

Google calendar renderes its weekLabel like above example.

Implementation idea:

```
This commit is a simple implementation that can parse weekLabel option with format string.

formatting rule:
- replace '{}' placeholder to week number

Given "第{}週" string in weekLabel option, it renders "第1週" text on weekLabel.

This code is tested with a limited unit test and rendering is confirmed by hand.

With its implementation, locale information file for Japanese language can be written like ja.ts.
```

### DIFF
--- a/packages/__tests__/src/legacy/weekLabel.ts
+++ b/packages/__tests__/src/legacy/weekLabel.ts
@@ -39,6 +39,14 @@ describe('weekText', () => { // TODO: rename file
         })
         expectWeekNumberTitle(calendar, 'YO')
       })
+
+      it('renders correctly when customized and given format string', () => {
+        let calendar = initCalendar({
+          direction: 'ltr',
+          weekText: '第{}週',
+        })
+        expectWeekNumberTitle(calendar, '第週')
+      })
     })
 
     function expectWeekNumberTitle(calendar, title) {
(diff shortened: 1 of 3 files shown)


hrkt:
weekLabel with this proposal:
image

current weekLabel implementation:
image


c.f. Google calendar:
image

hrkt:
Thank you for adding label, @acerix .

additional comment:

It might be specific for Japanese language, week number appears "in between" and "after" characters.
i.e.

"weekTextLong" option might be "第{number}週"
"weekText" option might be "{number}週"

Handling such case might be cumbersome, in such edge cases in i18n, it might be useful to give languages some degrees of formatting option.

In the scope of security consideration, if the i18n option will be given by programmers, replacing some characters will not cause a problem.
If this option might be given by external entities other than the site, we'll have to give some kind of sanitization logic here.

hrkt:
I also tried the displayed "week" label in other languages .

Chinese
image

Korean
image

hrkt:
merged main branch

PR #7108, opened 2022-12-23
replace "évènement" in french locales

Gladear:
Updated French (and -ca and -ch) locale for a typo on the word "évènement".
While both are correct, "évènement" would be the new standard (since 1990).

Here's a article (in french) explaining all of that: https://www.lalanguefrancaise.com/orthographe/evenement-ou-evenement-orthographe


arshaw:
Released in [v6.1.1](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.1)

Issue #6957, opened 2022-08-26
Support luxon 3.0


719media:
### Checklist

Please mark these items with an [x]

  - [x] I've already searched through [existing tickets](https://fullcalendar.io/issues)
  - [x] Other people will find this feature useful

### Feature Description

Luxon 3.0 released, luxon 2.0 plugin should be modified to include 3.0 peerDep, functionality is same sans default setting

zavan:
Everything seems to be working fine with luxon v3 when installed with `npm i --force`, but it's annoying having to run it like this every time. The current plugin is awkwardly named `/luxon2`, so not sure a new one should be created or that one should be updated...

gazben:
Any update on this? Is there an official way of using luxon 3?

DywiTom:
Bump

krishna-koushik:
merge please 🙏🏽

DywiTom:
Bump²

arshaw:
I just released the [@fullcalendar/luxon3](https://www.npmjs.com/package/@fullcalendar/luxon3) package in [v6.1.8](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.8). I'll write docs tomorrow. It works exactly as you might expect.

PS- please ignore the embarrassingly wrong README on npm. I've already updated it and queued it for the next release.

PR #7315, opened 2023-05-03
Allow luxon3 in luxon2 package

gazben:
Allow luxon3 as a dependency to `@fullcalendar/luxon2`. Luxon3 has only one breaking change, but this should not break the `fullcalendar` integration: https://github.com/moment/luxon/blob/master/docs/upgrading.md#2x-to-30

This feature has been requested in this issue: https://github.com/fullcalendar/fullcalendar/issues/6957

### DIFF
--- a/packages/luxon2/package.json
+++ b/packages/luxon2/package.json
@@ -11,13 +11,13 @@
   "homepage": "https://fullcalendar.io/docs/luxon2",
   "peerDependencies": {
     "@fullcalendar/core": "~6.1.6",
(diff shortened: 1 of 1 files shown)


arshaw:
I've created an official `@fullcalendar/luxon3` package rather than extending the semver for luxon2. Released in [v6.1.8](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.8). Regardless, thanks for the help @gazben

Issue #7364, opened 2023-06-06
Combination of Timezone changer and ICS is not working


danielpeintner:
### Reduced Test Case

https://gist.github.com/danielpeintner/c2a744aa2c37574d7e2e16686e64f84d

### Do you understand that if a reduced test case is not provided, we will intentionally delay triaging of your ticket?

- [X] I understand

### Which connector are you using (React/Angular/etc)?

No connector (vanilla JS)

### Bug Description

Based on your timezone demo https://fullcalendar.io/docs/timeZone-demo (which works fine!) I wanted to create a similar calendar for ICS files.

Hence I *mostly* took the "timeZone-demo" you are providing and changing

`events: 'https://fullcalendar.io/api/demo-feeds/events.json',`

with

```
events: {
	url: 'test.ics',
	// url: 'https://www.w3.org/groups/ig/wot/calendar/export',
	format: 'ics'
},
```

besides adding the necessary libraries

```
  <script src="https://cdnjs.cloudflare.com/ajax/libs/ical.js/1.5.0/ical.min.js"></script>
  <script src="https://cdn.jsdelivr.net/npm/@fullcalendar/icalendar@6.1.8/index.global.min.js"></script>
```

While in the case of feeding events.json to the calendar switching works fine... the same does *NOT* work for ICS event input.

NOTE: the [gist here](https://gist.github.com/danielpeintner/c2a744aa2c37574d7e2e16686e64f84d) contains the data to reproduce the issue. The file test.ics needs to be put on the same resource and served via HTTP server to avoid CORS issue.
Online test can be found here: http://peintner.org/tmp/fullcalendar-ics/index.html
It shows the calendar entries but changing the timezone in the upper/left corner does not work.

acerix:
The `timeZone` setting in that demo works by including the timezone in the URL params like "events.json?start=2023-05-28T00%3A00%3A00&end=2023-07-09T00%3A00%3A00&timeZone=Africa%2FBrazzaville" so the server returns the event data in the specified zone.  That param is not included for ICS feeds, although it seems the same would not work in general there, the current behaviour as in your demo does seem like a bug.  It also sometimes flickers incorrect times when changing zones.

danielpeintner:
Thanks for your update. Do you see a "simple" workaround that could be used instead.
e.g., parsing ICS file upfront, filter start/end range and give it then to fullcalendar like in the case with events.json input?

acerix:
[...]

PR #7450, opened 2023-09-09
Fix icalendar timezones

### LINKED ISSUE #7456: ICalendar events with 'TZID' displayed in wrong timezone
### Reduced Test Case

https://github.com/leoleoasd/calendar_bug_demo

### Do you understand that if a reduced test case is not provided, we will intentionally delay triaging of your ticket?

- [X] I understand

### Which connector are you using (React/Angular/etc)?

No connector (vanilla JS)

### Bug Description

reduced ical:
```
BEGIN:VEVENT
CREATED:20230917T160840Z
LAST-MODIFIED:20230917T171829Z
DTSTAMP:20230917T171829Z
UID:40597b6c-e9bf-4206-9a2f-97acc0b94f8f
SUMMARY:test
X-MOZ-LASTACK:20230917T162602Z
DTSTART;TZID=Asia/Omsk:20230917T100900
DTEND;TZID=America/New_York:20230917T010900
TRANSP:OPAQUE
X-MOZ-GENERATION:5
SEQUENCE:2
BEGIN:VALARM
ACTION:DISPLAY
TRIGGER:-PT15M
DESCRIPTION:Default Mozilla Description
END:VALARM
END:VEVENT
```
this event is shown as it begins on 10:09 am and ends on 11:09 am NO MATTER BROWSER LOCAL TIMEZONE SETTING:
image_2023-09-16_15-05-15

image_2023-09-16_15-00-19

Which should be (after fix in #7450):
image_2023-09-16_15-06-11

leoleoasd:
fixes #7456

Currently, icalendar module will treat all times as local timezones. Also, `TZID` in `DTSTART` of icalendar events are ignored, resulting more confusing result (as events with local timezones are displayed correctly but events with other TZID are shifted).
This PR fixes both these problems.
1. use `toJSDate().toISOString()` as `ical.js`'s `toString()` method don't include timezones other than UTC or local:

https://github.com/kewisch/ical.js/blob/b8157d34d61ce15843ea55cd05d9028414ccfdff/lib/ical/time.js#L1055-L1071

2. add timezone info to ical.js


acerix:
Please submit this ticket under Issues with a [runnable, stripped-down demonstration](http://fullcalendar.io/wiki/Reduced-Test-Cases/) of the bug, then link this PR to it.  I found a similar issue https://github.com/fullcalendar/fullcalendar/issues/7364 but it sounds like not the same as what this is fixing.

leoleoasd:
I discovered my problem and write this fix, this MAY fix #7364 as well but I havn't tested.

leoleoasd:
Before:
(Event should start at 12:00 and end at 13:00 in china timezone; I used devtools to change the timezone to china)
image_2023-09-16_15-00-19
(Event should start at 00:09 and end at 01:09 in ET; No devtool option used so it should my local TZ which is UTC-4)
image_2023-09-16_15-05-15
After:
image_2023-09-16_15-06-11

leoleoasd:
Any update? @acerix  @arshaw

koryhutchison:
I can confirm that this fixed my problem as well! Although it did weird things when I set a time zone on the calendar itself. But in local mode, it worked flawlessly. I needed this for airline flight events that are in my calendar that cross time zones. Would be great if this could get added officially to the repository.

PR #7471, opened 2023-10-03
fix translation 'list' in NL

WilcoSp:
`List` in Dutch isn't `Agenda` but `Lijst`
`Agenda` means `Calendar`/`Agenda` when translated into English from Dutch


arshaw:
Released in [v6.1.10](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.10). Thanks!

PR #7482, opened 2023-10-10
Draft: Extra cell with buttons in list view

kSzajo:
Hello,

I'd like to enhance list view with extra cell containing buttons that could be used to manipulate event. This would require to allow passing  optional `HTMLElement` (or some other interface to be framework agnostic ?) which then could be rendered in extra cell. Please refer to screenshot attached

Screenshot 2023-10-10 at 20 46 57

Could you please assist on how this idea can be developed? I did some initial draft but got really confused how to make it agnostic and what should be the API to allow rendering custom components in extra table cell.

I assume this requires change in event struct https://github.com/fullcalendar/fullcalendar/blob/main/packages/core/src/structs/event-source-parse.ts#L34 to allow pass `HTMLElement`?

### DIFF
--- a/packages/core/src/common/EventContainer.tsx
+++ b/packages/core/src/common/EventContainer.tsx
@@ -26,6 +26,7 @@ export interface MinimalEventProps {
 
 export type EventContainerProps = ElProps & MinimalEventProps & {
   defaultGenerator: (renderProps: EventContentArg) => ComponentChild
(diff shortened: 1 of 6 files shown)


acerix:
Please submit this ticket under Issues explaining the new feature you are looking for, then link this PR to it.

PR #7494, opened 2023-10-16
Update bg.ts

Icaka1337:
Updated Bulgarian translation in bg.ts


arshaw:
hi @Icaka1337 , there's a broken reference to `affix()`

Issue #7493, opened 2023-10-16
BG localization, week number is not correct


Ilchev:
### Reduced Test Case

https://github.com/fullcalendar/fullcalendar/blob/main/packages/core/src/locales/bg.ts

### Do you understand that if a reduced test case is not provided, we will intentionally delay triaging of your ticket?

- [X] I understand

### Which connector are you using (React/Angular/etc)?

Angular

### Bug Description

In BG localization, the week number is not correct:

current code is:
  week: {
    dow: 1, // Monday is the first day of the week.
    doy: 7, // The week that contains Jan 1st is the first week of the year.
  },

correct should be:
  week: {
    dow: 1, // Monday is the first day of the week.
    doy: 4, // The week that contains Jan 4th is the first week of the year.
  },

for for reference
https://www.timeanddate.com/calendar/?year=2023&country=73
"Week numbers: ISO 8601 (week starts Monday) - week 1 is the first week with Thursday."

arshaw:
released [v6.1.10](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.10)

PR #7495, opened 2023-10-16
Update bg.ts

Ilchev:
Bulgaria uses standard ISO 8601

#7493


arshaw:
Released in [v6.1.10](https://github.com/fullcalendar/fullcalendar/releases/tag/v6.1.10). Thanks!

Issue #7431, opened 2023-08-21
Provide drag start and stop events/callbacks for external draggable


joe-at-shipwell:
### Checklist

- [X] I've already searched through [existing tickets](https://github.com/fullcalendar/fullcalendar/issues)
- [X] Other people will find this feature useful

### Is this feature for a specific connector (React/Angular/etc)?

No connector in particular

### Feature Description

Our app needs to implement a feature where "unscheduled" appointments can be dragged onto the calendar to schedule them. We need to render background events indicating available time ranges when an appointment is picked up. We can do this for rescheduling appointments already on the calendar easily because there are callbacks. We can not do this for externally sourced appointments because callbacks are not provided. We can hack start by acting when the event creation callback is invoked, but we can not tell if the user changed their mind and dropped the appointment elsewhere, so we won't know to hide the availability window background appointments. 

Please, expose drag start and drag stop events or callbacks for external draggables, similar to how the eventDragStart and
eventDragStop callback properties work for rescheduling events within a single calendar.

DeveloperMarius:
Hello,

I created a PR to add this feature #7542.

~ Marius

Lazerproof:
Where useful addition, would be great to have it in package.

PR #7542, opened 2023-11-29
Provide drag start and end events for external draggable

DeveloperMarius:
Hello @arshaw,

I implemented support for the drag start and end events for external draggable elements (#7431).
My use case is that I have a table with a different layout then the fc-event elements should have. I want to allow users to drag the table rows into the calender. This feature allows me to manipulate the mirror element after the drag start to match the correct fc-event styling.

An option to generate the mirror element would go to far or would that be useful too?

I did not compile the code oder added the options to the documentation.
I'm looking forward to your feedback.

~ Marius

### DIFF
--- a/packages/interaction/src/interactions-external/ExternalDraggable.ts
+++ b/packages/interaction/src/interactions-external/ExternalDraggable.ts
@@ -1,9 +1,11 @@
 import { BASE_OPTION_DEFAULTS, PointerDragEvent } from '@fullcalendar/core/internal'
 import { FeaturefulElementDragging } from '../dnd/FeaturefulElementDragging.js'
-import { ExternalElementDragging, DragMetaGenerator } from './ExternalElementDragging.js'
+import { ExternalElementDragging, DragMetaGenerator, DragEventStart, DragEventEnd } from './ExternalElementDragging.js'
 
 export interface ExternalDraggableSettings {
   eventData?: DragMetaGenerator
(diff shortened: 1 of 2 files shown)


Lazerproof:
Where useful addition, do you plan to include it to  the package?

Lazerproof:
In case anyone is looking for a solution, here's how I solved in my case

```
const drag = new Draggable(unscheduledRef.value, {
        itemSelector: '.is-unscheduled-item',
        eventData: function (eventEl) {
            ....
        },
    })

    drag.dragging.emitter.on('dragstart', function (e) {
        isDragging.value = true
    })

    drag.dragging.emitter.on('dragend', (e) => {
        if (isDragging.value && hoveredEvent.value?.id && hoveredEvent.value?.extendedProps?.type === TYPE_SESSION) {
            const subjectEl = e.subjectEl

            const dataAttributeValue= subjectEl.getAttribute('data-event')
            const parsedData= JSON.parse(dataAttributeValue)
            console.log(parsedData)
        }

        nextTick(() => {
            isDragging.value = false
        })
    })
```

DeveloperMarius:
@arshaw is this an interesting addition or should I just implement my workaround in my project and discontinue this PR?
What would I have to do, to make this work for you?

~ Marius

DeveloperMarius:
Hello @arshaw,

I would appreciate a short feedback regarding my last message.
Thank you.

 ~ Marius

Issue #7492, opened 2023-10-15
Bug: can't use ThirdPartyDraggable to drag SortableJs element into FullCalendar


albertpratomo:
### Reduced Test Case

https://stackblitz.com/edit/github-j6cvuf?file=src%2FDemoApp.vue

### Do you understand that if a reduced test case is not provided, we will intentionally delay triaging of your ticket?

- [X] I understand

### Which connector are you using (React/Angular/etc)?

Vue

### Bug Description

I have a list of events element that is sortable (using [SortableJS](https://github.com/SortableJS/Sortable)).

I want user to be able to drag an event element, into FullCalendar, then an event should be created.

I can use `Draggable` from '@fullcalendar/interaction' for this, but then the events element won't be sortable (SortableJS doesn't work).

Unfortunately if I use `ThirdPartyDraggable`, SortableJS itself work, but the event is not draggable into FullCalendar.

How can `ThirdPartyDraggable` be used with SortableJS?

Many thanks.

acerix:
It seems to work if the calendar is clicked while dragging the events,  but it should drag in without needing an extra click so it seems a bug that SortableJS is not compatible.  I believe using another lib like Dragula which allows sorting would not have this issue.

albertpratomo:
Thanks for confirming the bug @acerix. This will be fixed right? SortableJS is one of the most common sorting library out there 😃

buditanrim:
@acerix Hey, just curious if there's any update on this?

amommersteeg:
I believe I was able to get it to work the way you wanted, by using `forceFallback: true`. I updated your example in the link above with the following:
```
useSortable(el, list, {
  forceFallback: true,
});
```

This solution was inspired by this ticket, https://github.com/SortableJS/Sortable/issues/1268.

PR #7588, opened 2024-01-30
Pure HTML drag event handling

### LINKED ISSUE #7593: ThirdPartyDraggable does not handle HTML drag events
### Reduced Test Case

https://codepen.io/pducas-theatrixx/pen/KKEeBKv

### Do you understand that if a reduced test case is not provided, we will intentionally delay triaging of your ticket?

- [X] I understand

### Which connector are you using (React/Angular/etc)?

No connector (vanilla JS)

### Bug Description

I work on an application using a custom drag and drop handler based on pure HTML dragging (draggable elements are tagged with the [`draggable`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable) attribute). Here is a link to [HTML Drag and Drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) for reference.

We used the `ThirdPartyDraggable` class to integrate FullCalendar into this dragging system, but it does not work as intended. When an element is dragged over or dropped onto the calendar, it is not acknowledged by the view. However, if you then move he mouse and click on the calendar, an event will be created. This behaviour is the same in month, week and day grids.

I looked into the code, and I think it is linked to the `PointerDragging` class only listening to `mousemove` and `mouseup` to handle drag and drop events. When an element has the `"draggable"="true"` tag, these events are replaced respectively by `dragover` and `dragend`.

FeelDee:
Fixes #7593

As of now, the `PointerDragging` class only listens to `mousemove` and `mouseup` to handle dragging events. However, when an element has the `"draggable"="true"` tag, these events are replaced respectively by `dragover` and `dragend`.

See [HTML Drag and Drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) for reference.

This PR adds new listeners for these events, allowing to track both methods of dragging without any further configuration needed. This might not be the best method and might need some improvements for edge cases, but it's a working one.

### DIFF
--- a/packages/interaction/src/dnd/PointerDragging.ts
+++ b/packages/interaction/src/dnd/PointerDragging.ts
@@ -103,9 +103,11 @@ export class PointerDragging {
 
       if (!this.shouldIgnoreMove) {
         document.addEventListener('mousemove', this.handleMouseMove)
+        document.addEventListener('dragover', this.handleMouseMove)
       }
 
       document.addEventListener('mouseup', this.handleMouseUp)
+      document.addEventListener('dragend', this.handleMouseUp)
     }
   }
 
@@ -117,7 +119,9 @@ export class PointerDragging {
 
   handleMouseUp = (ev: MouseEvent) => {
     document.removeEventListener('mousemove', this.handleMouseMove)
+    document.removeEventListener('dragover', this.handleMouseMove)
     document.removeEventListener('mouseup', this.handleMouseUp)
+    document.removeEventListener('dragend', this.handleMouseUp)
 
     this.emitter.trigger('pointerup', this.createEventFromMouse(ev))
 


FeelDee:
Created this PR by accident by incorrectly merging it into the original repo. I'm reopening it now that a complete bug report has been submitted.

acerix:
Fixes #7492

arshaw:
@FeelDee, thank you so much for this PR. I bet this solution works, but I'm a bit apprehensive to start listening to drag-related events when the class is already so heavily oriented towards explicit mouse/touch events. I'd be afraid the three might not play well. A lot more manual testing would need to be done. Tbh, I'll probably refactor fullcalendar's DnD system for the next major version b/c I'll attempt to support React Native.

In the meantime, [this workaround](https://github.com/fullcalendar/fullcalendar/issues/7492#issuecomment-1871225484) sound like it'll get the job done.

FeelDee:
@arshaw thank you for your detailed answer, we'd really like this issue to be fixed so your commitment is very much appreciated. I had the same apprehension regarding unwanted behaviour since we are directly meddling with core events, yet with my little knowledge on the fullcalendar code base I couldn't think of any better solution.

Sadly, the workaround proposed in the bug report does no fix our problem, because we use a custom implementation based on the HTML Drag and Drop API that does not rely on Sortable JS. Changing this behaviour would imply some HTML dark magic that we'd really rather not use.

While waiting for the next major version, is there any way we could patch and deploy a custom version of the interaction package? I had a lot of trouble having the postinstall script to work in production, so any doc or hint would really help.

michaelfaith:
Hi there. Just wondered if any help is needed to move this forward.  Happy to contribute, if needed.

PR #7697, opened 2024-06-12
Update Popover.tsx

Kundros:
I fixed bug of popover being sometimes off boundry. Firstly when table is scrollable on x axis it wasn't taken in account. Secondly, the calculation for border was sometimes off, I am not sure how it was inteded to work but I changed it a bit so now it is calculated from parent width and I removed padding. You might want to double check my solution, not sure if it is universal but imo it should be. I needed to make that change for my usecase at least.


acerix:

Issue #6150, opened 2021-02-07
Add documentation for event.mutation


jrock2004:
### Checklist

Please mark these items with an [x]

  - [x] I've already searched through [existing tickets](https://fullcalendar.io/issues)
  - [x] Other people will find this feature useful

### Feature Description

Please describe what this feature will do.

currently there is no way to set multiple event properties at once and you have to do them one by one. The issue is that is every time you use a setProp or setExtendedProp it will call the event change function which in turn might call an API to update a database. So it looks like there is an event.mutation that will allow you to pass in multiple properties at once. I looked through the documentation and I do not see it documented so not sure if it’s safe to use. 

If there is a better solve for this?

acerix:
I'm not sure how calling `setProp` would trigger an update to the database,  could you not user multiple `setProp` then only make one API call separately?  Another option is to make the update in the database first, then call `refetchEvents` to update all event data.  Using undocumentated private functionality is not recommended.

jrock2004:
@acerix Well when you call `setProp` or any of the sets, it will call the `eventChange` which in there is when I call the API to update the DB. Maybe there is a better hook I should be doing that in?

acerix:
I can see how it would be useful to be able to update multiple event props in a batch so that `eventChange` only get called once.   

For now, instead of using that hook you could just make the API call right after all setting all the new properties,  and use `eventDrop` if you also want an API call when an event is dragged, for example.

rome-legacy:
hey amigos,
first of all, thx for the great calendar! 
i have the same problem. would be great if there would be a possibility  of setting multiple event fields with one eventchange callback for DB update.
currently i'm opening a dialog on event click and provide some input fields for event modifications. then on btn-save click i'm setting properties on the event which calls eventChange multiple times.

for now i will switch to updating api when btn-save was clicked instead of eventChange callback but it has a small drawback: if some of the event.set... method calls fails, you would have inconsistent data in the database, since revert() might be only half of the truth.

regards

montulli:
I had a similar problem.  I needed to to update the properties of several hundred events, but the setProp call does a rerender on each call.   That made the total time to update all the events super slow.   I fixed it by setting the global rerenderDelay to 100ms and the rerenders stopped causing a problem for that use case.   I would suggest a simple fix would be to add a 'render' boolean to the setProp flag that allows the caller to disable the rerender of the event.

Destabilizator:
[...]

PR #7718, opened 2024-07-04
Silent Updated for props and extended props

sehaj23:
Giving an option to update the props and extended props in a silent manner without invoking eventChange function

Issue- https://github.com/fullcalendar/fullcalendar/issues/6150


NTICompass:
I think the `silent` parameter should also be added to `setStart`, `setEnd` and `setDates`.

PR #7755, opened 2024-09-07
Update event-source-def.ts

happybanana78:
fixed issue where fullcalendar would only load "description" and "location" inside the extendedProps attribute.


acerix:

PR #7788, opened 2024-10-07
Just return "Terminübersicht" in viewHint()

kozi:
The *glue* with the suffix `ansicht` does not work for `Terminübersicht`.


acerix:

kozi:
refs #7791

PR #7840, opened 2024-12-13
Update sr localization to be latin - sr-cyrl already exists for cyrilic

fl1k:
Serbian language uses both Latin and Cyrilic script - Since sr-cyrl already exists for cyrilic script sr should be latin - so update sr to use latin letters for system localization


vlazic:
I also came across this issue and I wanted to create an identical PR, so I suggest you go ahead and merge this PR!

Issue #7848, opened 2024-12-19
Unable to Prevent Specific Events from Being Resized from the Start


baranhr:
### Reduced Test Case

https://codesandbox.io/p/sandbox/jzdtkc

### Do you understand that if a reduced test case is not provided, we will intentionally delay triaging of your ticket?

- [X] I understand

### Which connector are you using (React/Angular/etc)?

React

### Bug Description

**Description**
I want to prevent specific events on the calendar from being resized to the left. However, I’ve encountered some issues with the existing options:

- The **durationEditable** property disables resizing on both sides of the event.
- The **eventResizableFromStart** property disables resizing from the start but applies globally to all events.

What I need is a way to control this behavior for specific events based on a condition in the event object.

For example:

- Some events should be resizable from the start.
- Others should not be resizable from the start.

Currently, I attempted to use the **startEditable** property in the event object to achieve this, but it doesn’t work as expected. Even when:

durationEditable: true
editable: true
resourceEditable: true
...and I explicitly set **startEditable: false**, the event is still resizable from the start.

**Steps to Reproduce**

1. Configure an event object with the following properties:
`{
  durationEditable: true,
  editable: true,
  resourceEditable: true,
  startEditable: false,
}`
2. Attempt to resize the event from the start.

**Expected Behavior**
The event should not be resizable from the start when **startEditable** is set to **false**.

**Actual Behavior**
The event remains resizable from the start despite **startEditable: false**.

**Questions**

- Is there an issue with how startEditable works?
- Am I using the property incorrectly, or is there another way to achieve this functionality?

baranhr:
I opened a pull request for this issue. I test it and it works. Can you review and test ?  https://github.com/fullcalendar/fullcalendar/pull/7849

barangezen:
I forked the project, applied a fix, and registered the package using Verdaccio. I then tested it in different projects, confirming that it worked successfully.

acerix:
Please refer to the [support page](https://fullcalendar.io/support/) and use [Stack Overflow](https://stackoverflow.com/questions/tagged/fullcalendar) for help.  If this is a bug, please supply a [runnable, stripped-down demonstration](http://fullcalendar.io/wiki/Reduced-Test-Cases/).

I believe the issue is with setting `editable: true` as that is a shortcut which sets both `startEditable` and `durationEditable`.

PR #7849, opened 2024-12-20
fix: make each event individually resizable at the start

baranhr:


baranhr:
For https://github.com/fullcalendar/fullcalendar/issues/7848

PR #7868, opened 2025-01-21
feature: insert styleSheet when switching between micro-applications after keep-alive is enabled

githubOfLiLin:
The styleSheet in style[data-fullcalendar] will be lost when switching between micro-applications after keep-alive is enabled。
So the styleSheet should be reinserted when styleEl.sheet.cssRules.length is empty


acerix:
Please submit a ticket under Issues describing this feature request, then link this PR to it.

githubOfLiLin:
> Please submit a ticket under Issues describing this feature request, then link this PR to it.

https://github.com/fullcalendar/fullcalendar/issues/7886

Issue #7860, opened 2025-01-09
Support ICS `COLOR`property


elijaholmos:
### Checklist

- [X] I've already searched through [existing tickets](https://github.com/fullcalendar/fullcalendar/issues)
- [X] Other people will find this feature useful

### Is this feature for a specific connector (React/Angular/etc)?

No connector in particular

### Feature Description

The [ICS RFC § 5.9](https://datatracker.ietf.org/doc/html/rfc7986#section-5.9) indicates that a `COLOR` property can be specified for iCal events. Looking at the current fullcalendar code, it seems that this property is not extracted when parsing ICS event data:

https://github.com/fullcalendar/fullcalendar/blob/9a0edaeb5eddf2cec8bd2409ecc26c4598b98a48/packages/icalendar/src/event-source-def.ts#L105-115

The dependent library, https://github.com/kewisch/ical.js supports the ICS `COLOR` property as of [release v1.5.0](https://github.com/kewisch/ical.js/releases/tag/v1.5.0). This version is compatible with the currently specified peer dependency of `^1.4.0`. 

I would be comfortable creating a pull request for this feature if the maintaining team is willing to approve it.

captn3m0:
I went ahead and filed https://github.com/fullcalendar/fullcalendar/pull/7923

PR #7923, opened 2025-04-20
[icalendar] Adds support for color

captn3m0:
Same as https://github.com/fullcalendar/fullcalendar/issues/7860

I'm maintaining this one-line patch in-tree at blr.today for the last year, and it hasn't caused any issues.


balsoft:
I would find this really useful too.

PR #7933, opened 2025-05-04
Fix/2542 dragging event to edge

ceddup:
Available options :

- dragToEdges: Boolean, defaults to false. Navigate to previous/next view when dragging an event to an edge.

- dragToEdgesFlashTitle: Boolean, defaults to true. When navigating to next/previous view - while event dragging - flash the title, current month tipically, so the user gets a visual confirmation of his action. Do override the "fc-flash-title" css class on your page to customize it.

- dragToEdgesShowSideBars: Boolean, defaults to true. Show a vertical blue bar on the edge when the view is about to get swapped. Do override the "fc-edge-nav-next" and "fc-edge-nav-prev" classes on your page to customize these.

- dragToEdgesOnMobileOnly: Boolean, defaults to false. Enable "dragToEdges" option on mobiles only.

- dragToEdgesOnDesktopOnly: Boolean, defaults to false. Enable "dragToEdges" option on desktops only.


ignatiusperera:
please merge this <3

PR #8011, opened 2025-11-03
Add daygrid-overlaps plugin with proportional multi-day layout

adrian0015:
## Summary
- forked the daygrid plugin into packages/daygrid-overlaps with its own package name and globals, keeping the existing API surface
- updated computeFgSegPlacement to let equal-ordered events overlap and to size multi-day slices proportionally when constrained by content height
- passed event order specs into TableRow and added an integration test showing a 10:00 end no longer conflicts with a noon start

## Testing
- pnpm -w build

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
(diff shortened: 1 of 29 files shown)


linonetwo:
What does this looks like?

PR #8035, opened 2026-02-24
fix(ui): truncate long titles, popover overflow

alirezaalavi87:
- Opening the "more" popover on days with many events (say, 100) would make the whole page scroll. Now the popover scrolls within itself
- Events with long titles would cause the popover to expand and scroll the whole page horizontally. now the popover has a max-width and long titles get truncated properly.

image

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,4 +1,5 @@
 engine-strict = true
-use-node-version = 20.19.0
+use-node-version = 24.15.0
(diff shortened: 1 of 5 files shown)


alirezaalavi87:
@acerix  Hi.
I think this is a pretty good fix. I don't want it to go stale. can you take a look at it and merge it if it's ok?

